PR #86, opened 2024-04-13
Rename EIDSCA docs folder to fix build script on unix

fflaten:
Update-EidscaTests.ps1 defaults to uppercase folder-name for EIDSCA docs.
https://github.com/maester365/maester/blob/ffe36539a86170c287ebf1414bd97010f14381d9/build/EIDSCA/Update-EidscaTests.ps1#L17-L18

Renamed folder to make it work on case sensitive filesystems like UNIX.


Cloud-Architekt:
Thanks for the PR! I'm currently working on the build script to use lower-case paths only for all EIDSCA content which seems to fits also to the overall folder structure in maester.

PR #122, opened 2024-04-17
Add skipScopeCheck param

rvdwegen:
Useful when your app has the needed effective access through other scopes.


rvdwegen:
need to rethink this

PR #124, opened 2024-04-22
Addition of CISA ScubaGear Section 2

Snozzberries:
Addition of section 2 for ScubaGear controls around risk based policies.

This required an additional scope for identity protection endpoints.


Snozzberries:
I also left the functions out of the PSD1 file, but noticed the GH Actions are failing on that Pester test. Let me know if on future PRs you'd want me to update that as well.

PR #147, opened 2024-04-26
Update Invoke-Maester.ps1

dnsinit:
Updated MailRecipient example description.


merill:
Good catch. Thanks for the PR @dnsinit 🙏

PR #156, opened 2024-05-04
Add CISA SCuBA Section 5

Snozzberries:
The `AuthorizationPolicy` and `ConsentPolicy` object types. I still don't think I full understand the intent around those, so that may need tinkering.

Two cmdlets use the same relative URI, not sure at what point a new "Get-Mt*" cmdlet is beneficial.
`$result = Invoke-MtGraphRequest -RelativeUri "policies/authorizationPolicy" -ApiVersion v1.0`


Snozzberries:
Added password policy check aligned with the CISA Rego, but there are alternative paths that may require validation for the intent of the control to be met.

merill (approved):

merill:
Wow, that's a ton of new tests. Amazing work @Snozzberries !

PR #157, opened 2024-05-07
Update EIDSCA.AG01.md

dnsinit:
Updated with new dates for deprecation to reflect what's shown in Entra portal.


merill (approved):
Good catch. Thanks for the PR @dnsinit

Issue #162, opened 2024-05-08
Feature request - Be able to run Invoke-Maester and not force connection to GraphAPI


Luunk:
The framework and report is so nicely done and useful. It would be nice to be able to run Invoke-Maester and not be forced to connect to GraphAPI if your custom tests does not need to use the GraphAPI. 
The idea being that you can write your own custom test for pretty much anything, have some markdown files prepared for each test and then generate the report with Maester. 
Like if there would be a parameter called "-Local" or something you could use on the Invoke-Maester comandlet. And then Invoke-Maester would not connect to GraphAPI and only run the local custom test provided. 

I know the whole point of Master is to test and validate EntraID with GraphAPI. But it's starting to be such a good pester report framework it would be nice to be able to use it for other tests as well that is not reliant on GraphAPI. But still get the same report layout. 

OK to close if it's out of scope, just a thought we had internally when testing to implement Maester. ☺️

merill:
I love the idea of using just the Maester framework to run custom tests that don't call graph.

This use case never crossed my mind.

It's definitely something we should add. I'll add this to the backlo.g

merill:
I propose calling the parameter **-SkipGraphConnect**

Luunk:
Yeah, sound like a good name for the parameter and what it is intended for 👍

DrPye:
Hi Guys, ive made the following PR #163 to address this. Hope its what you where after!

merill:
Thanks to @DrPye the option to -SkipGraphConnect is available now.

<img width="850" alt="image" src="https://github.com/maester365/maester/assets/1288081/9f615ee6-48fd-47e2-a3b9-045ec74d07b4">

PR #163, opened 2024-05-10
MSGraph connection check bypass

DrPye:
This adds the option to bypass the graph connection per #162 

It simply adds an extra switch param of ```SkipGraphConnect``` to ```invoke-maester``` and then puts the ```Test-MtContext``` behind an if statement. Tested locally.

I've also updated the cmdlets document too.

Please let me know if you expected something else


merill:
Awesome. Thanks for the PR. @DrPye 

This looks good. In future the cmdlet docs don't need to be updated manually since they are auto-generated based on the help.

PR #167, opened 2024-05-10
Use allow-no-subscriptions makes it easier

svrooij:
They explicitly made an option if you want to login to the cli and have no access to any subscription. Let's use this option and make configuring it for github even easier.

https://github.com/Azure/login?tab=readme-ov-file#allow-no-subscriptions


merill:
Amazing! I didn't know about this. One of the best PRs. Thanks a lot @svrooij ❤️

Issue #169, opened 2024-05-12
Missing CISA SCuBA Controls


Snozzberries:
Currently Entra section 4 is not implemented as it requires Azure Management API integrations.

Currently Entra control 8.3 is not implemented as the Graph API does not expose the domain restrictions for B2B invitations. Alternative control validation could be used, but would take liberties when compared to CISA's implementation.

PR #170, opened 2024-05-12
SCuBA Section 7 & 8

Snozzberries:
This wraps up the Entra section. I created issue #169 for tracking of a few gaps in the implementation.


merill:
Wow. Nice work @Snozzberries 

Would it be possible to make one change please?

For the roles, can we switch to using the GUIDs instead of the display name. This way it won't break functionality if the display name is ever changed.

Plus it will be good to have a short description for the switch parameter in Get-MtRole

Snozzberries:
I think that should cover what you mentioned. Thanks for the feedback @merill.

merill:
Another top notch addition @Snozzberries 

Amazing work.

merill:
This has now been published to the PowerShell module. @Snozzberries quick note that I made some updates to show better message when the test passes/fails for https://github.com/maester365/maester/blob/main/powershell/public/CISA/Entra/Test-MtCisaGuestUserAccess.ps1

One was displaying the actual name of the role from the UI and the other is a deep link to the portal in the steps to make it easier to directly go to the blade.

Snozzberries:
Thanks @merill still feeling out where to take liberties, appreciate the guidance and feedback.

PR #182, opened 2024-05-18
Swapped to ParamSets to make it more maintainable than throwing when …

Snozzberries:
…necessary params aren't set.

Hacked this one together when I was first completing this. Wanted to go back and make it a little more maintainable.


merill:
Cheers. Thanks Michael, this is definitely better.

PR #187, opened 2024-05-22
Get-MtUserAuthMethod

Arcath:
Adds a function that users can use in their own tests to get the MFA status of a user.

## Usage:
```ps1
> Get-MtMFAMethods -UserId <#Object-GUID#>
userId                               enabled methods
------                               ------- -------
<#Object-GUID#>                      True    {phone, password, microsoftAuthenticator, microsoftAuthenticator}
```

My use case is testing that all staff who have been assigned to a group that should have MFA setup do infact have MFA setup.


Arcath:
The return of all the methods on there would also allow for tests like "GAs must be using Microsoft Authenticator", "No user should be using phone authentication", for example.

merill:
@Arcath thanks a lot this is a good cmdlet to add.

Here are my review comments.

- Can we rename to Get-MtUserAuthMethod with a switch parameter to filter out to just MFA? Quick note that method name should be singular (ie not Get-...Methods instead Get-...Method)
- I wrote something recently for the MSID module, feel free to copy that over. We will need to keep it up-to-date as new methods are added. But Email is another one that is a single factor. See. https://github.com/AzureAD/MSIdentityTools/blob/b029c1c8277fe4b83b32e7aa8b7d899fdca88833/src/Export-MsIdAzureMfaReport.ps1#L378

merill (changes requested):
@Arcath thanks a lot this is a good cmdlet to add.

Here are my review comments.

Can we rename to Get-MtUserAuthMethod with a switch parameter to filter out to just MFA? Quick note that method name should be singular (ie not Get-...Methods instead Get-...Method)
I wrote something recently for the MSID module, feel free to copy that over. We will need to keep it up-to-date as new methods are added. But Email is another one that is a single factor. See. [AzureAD/MSIdentityTools@b029c1c/src/Export-MsIdAzureMfaReport.ps1#L378](https://github.com/AzureAD/MSIdentityTools/blob/b029c1c8277fe4b83b32e7aa8b7d899fdca88833/src/Export-MsIdAzureMfaReport.ps1#L378)

Arcath:
Couple of changes:

Renamed to `Get-MtUserAuthMethod`
 
Added `Exclude` which allows you to exclude an authentication method from the output. For Example (using the same user as before):

```ps1
> Get-MtUserAuthMethod -UserId <#Object-GUID#> -Exclude "phone", "password"
userId                               enabled methods
------                               ------- -------
<#Object-GUID#>                      True    {microsoftAuthenticator, microsoftAuthenticator}
```

Added `NonMFAMethods` for users to control how the `enabled` response works. This defaults to `@("password")`, could easily add `email` to that list as well. For Example:

```ps1
> Get-MtUserAuthMethod -UserId <#Object-GUID#> -NonMFAMethods "password", "phone", "microsoftAuthenticator"
userId                               enabled methods
------                               ------- -------
<#Object-GUID#>                      False   {phone, password, microsoftAuthenticator, microsoftAuthenticator}
```

f-bader:
@merill do we want to rely on `Get-MgUserAuthenticationMethod` or should we always use `Invoke-MtGraphResults`
This adds overhead but we don't have to rely on the current Graph modules too much?
Otherwise I think the PR looks good to merge.

Arcath:
@f-badger Good point, I hand't noticed that. I've swapped it to `Invoke-MtGraphRequest` now.

merill (approved):
@Arcath Thanks a lot for the updates and apologies for the delay.

I updated the method to return the original graph objects and appended the IsMfa and displayname to the objects. 

This will allow for wider use of the method for the various tests.

PR #190, opened 2024-05-23
Various CISA tests has been fixed

tdcthosc:
The following tests has been hotfixed by changing -All to -Active until a bug regarding read-only permissions for roleEligibilityScheduleRequests is fixed. See: https://learn.microsoft.com/en-us/answers/questions/1656176/unable-to-get-rolemanagement-directory-roleeligibi
* Test-MtCisaCloudGlobalAdmin.ps1
* Test-MtCisaGlobalAdminCount.ps1
* Test-MtCisaGlobalAdminRatio.ps1

Test-MtCisaPermanentRoleAssignment.ps1 also fails due to above bug. This has been fixed by adding API permission "RoleManagement.Read.All" to Microsoft Graph scopes.

A divide by zero condition in Test-MtCisaGlobalAdminRatio.ps1 has been fixed. This occurs if no users exists in any other privileged assignments outside Global Administrators.

An output mistake in the following tests has been fixed.
* Test-MtCisaAppUserConsent.ps1
* Test-MtCisaCloudGlobalAdmin.ps1
* Test-MtCisaUnmanagedRoleAssignment.ps1

Test-MtCisaAppRegistration.Tests.ps1 used the wrong function to test with.

Skip conditions has been moved to Describe statement instead for the following tests, otherwise it would produce undesired markdown descriptions from Test-MtCisaPhishResistant.
* Test-MtCisaAuthenticatorContext.Tests.ps1
* Test-MtCisaMfa.Tests.ps1

Adjusted scopes in Get-MtGraphScope
* 'RoleManagement.Read.All' has been added.
* 'Policy.Read.ConditionalAccess' has been removed. It should not be required when 'Policy.Read.All' is already part of the scopes.
* '/maester/website/docs/sections/permissions.md' has been updated accordingly.

Issue #76, opened 2024-04-13
Test-MtConditionalAccessWhatIf fails with response 'Forbidden'


jg-declarative:
Custom Conditional Access tests using the provided Test-MtConditionalAccessWhatIf function fail with forbidden response code.

Same error generated with GitHub Actions and local Invoke-Maester command.

Powershell version 7.4.1
User has Global Admin permissions on the tenant.

![image](https://github.com/maester365/maester/assets/165257172/848c104d-ceb1-4e02-86a5-2adad7f82353)

Attempted tests (userids removed):

`Describe "Conditional Access WhatIf"  {

        It "CAP.001: Block access to the Azure portal for non-admin users" {
            Test-MtConditionalAccessWhatIf  -UserId '<GUID>' -IncludeApplications "00000002-0000-0ff1-ce00-000000000000"  | Should -Contain "block"
        }

}
Describe "Contoso.ConditionalAccess" {
    It "Microsoft 365 access requires MFA" {

        $userId = (Get-MgUser -UserId 'user@domain.com').Id
        $sharePointAppId = '67ad5377-2d78-4ac2-a867-6300cda00e85'

        $policiesEnforced = Test-MtConditionalAccessWhatIf -UserId $userId `
            -IncludeApplications $sharePointAppId `

        $policiesEnforced.grantControls.builtInControls | Should -Contain "mfa"
    }
}`

fflaten:
It requires the `Policy.ReadWrite.ConditionalAccess` scope which isn't requested atm. AFAIK there's no read-only permission yet during the private preview of the underlying graph API. Likely one of the reasons they disabled the included tests for the moment (https://github.com/maester365/maester/commit/d548a5af03e89ecb179708c6ffab4c1c075a3d37)

f-bader:
@fflaten thanks for pointing this out. We still should add some documentation here.
@jg-declarative If you connect to graph using `Connect-MgGraph -Scopes "Policy.ReadWrite.ConditionalAccess"` one time you can grant this permission for your users.

jg-declarative:
@fflaten @f-bader thanks guys super helpful! Are these still undocumented from the Microsoft side?

f-bader:
Yes public documentation should arrive soon. Thanks to the very gracious product manager we got approval to use it a bit early.

RobinDadswell:
> It requires the `Policy.ReadWrite.ConditionalAccess` scope which isn't requested atm.
> 
[...]

PR #199, opened 2024-05-28
Docs: Add 'Policy.Read.ConditionalAccess' scope

tbalthazar:
## Context

I'm currently setting up Maester to run in a DevOps pipeline.
I followed [the documentation](https://maester.dev/docs/monitoring/azure-devops/#grant-permissions-to-microsoft-graph) to grant permissions to the Microsoft Graph and got this error:
```
     | These Graph permissions are missing in the current connection =>
     | (Policy.Read.ConditionalAccess). Add the missing 'Application'
     | permissions in the Microsoft Entra portal and grant consent. You will
     | also need to Disconnect-Graph to refresh the permissions
```

I added the `Policy.Read.ConditionalAccess` permission and it worked.

After searching for related issues here, I saw the `Get-MtGraphScope` function had recently been updated to add this permission, but AFAICT the documentation hasn't been updated yet: https://github.com/maester365/maester/pull/178

## Proposed Changes

- Update the documentation to reflect the updated permissions required for the Microsoft Graph app.

I hope this helps!

### DIFF
--- a/website/docs/sections/permissions.md
+++ b/website/docs/sections/permissions.md
@@ -5,3 +5,4 @@
 - **PrivilegedAccess.Read.AzureAD**
 - **IdentityRiskEvent.Read.All**
 - **RoleEligibilitySchedule.Read.Directory**
+- **Policy.Read.ConditionalAccess**


Snozzberries:
The change proposed in #190 would address this as well. As `Policy.Read.ConditionalAccess` is included in `Policy.Read.All`.

tbalthazar:
> As Policy.Read.ConditionalAccess is included in Policy.Read.All.

@Snozzberries That's what I would have expected as well. But not explicitly adding `Policy.Read.ConditionalAccess` currently doesn't work for me. I haven't tried to add `RoleManagement.Read.All` as done in #190 though.

Snozzberries:
Hmm, my only thought is some nuance between interactive and application permission delegations.

If you run `Connect-MgGraph -Scopes Policy.Read.All` and then `Get-MtConditionalAccessPolicy` does that throw any error in your tenant when run interactively?

If it does then can you try  `Connect-MgGraph -Scopes Policy.Read.All` and then `Get-MtConditionalAccessPolicy` to see if that resolves it.

If it doesn't throw any error then it seems related to application delegation permissions and I would need to see if I could recreate.

tbalthazar:
Unfortunately my account doesn't have the appropriate permissions in our tenant for me to test this :/

Snozzberries:
> Unfortunately my account doesn't have the appropriate permissions in our tenant for me to test this :/

Sounds like the role you have in your tenant may be authorized to use `Policy.Read.ConditionalAccess` but not `Policy.Read.All`. 

Is it possible to verify those delegations or try an elevated role in your tenant to confirm?

Snozzberries:
With the following `Get-MtConditionalAccess` is able to return results.

image

I actually had the issue of until I added `Policy.Read.All` the cmdlet would return HTTP 403.

image

tbalthazar:
> Sounds like the role you have in your tenant may be authorized to use Policy.Read.ConditionalAccess but not Policy.Read.All.

The service principal originally got those permissions granted:
image

I confirm it did not work with just `Policy.Read.All` and it started working after I added `Policy.Read.ConditionalAccess`.

Snozzberries:
@f-bader #202, #190, and this are referencing the addition of `Policy.Read.ConditionalAccess`. It looks like this was added to scopes in #76 to support WhatIf.

I was able to confirm the `Policy.Read.ConditionalAccess` is **not** needed for `Invoke-Maester -SkipGraphConnect` today, as that bypasses the `Test-MtGraphContext` check where that scope is checked and an error is thrown if the SPN was configured with the permissions listed in the (docs)[https://maester.dev/docs/monitoring/github#grant-permissions-to-microsoft-graph].

Do you want to add switch logic for making `Policy.Read.ConditionalAccess` optional or add it to the docs as an additional required permission?

f-bader:
I guess I will do both. 
Add a check / switch in the code like the mail permission
Add it as optional requirement in the docs

Can do the code and docs changes mid next week

merill:
This is going to be needed by the WhatIf API as an explicit permission (it might not be included in the Policy.Read.All for some time - or ever...). Given that whatif is one of the main use cases and it will be used in future tests it will be better to consent for this up-front.

So I'm going to accept this PR and merge it in.

Thanks a lot for raising this and the discussion folks. Much appreciated.

Issue #194, opened 2024-05-27
MS.AAD.3.7 and MS.AAD.3.8 assume you're running Hybrid


JonesMikael:
I have som suggestions on the following tests:
```
MS.AAD.3.7: Managed devices SHOULD be required for authentication.
MS.AAD.3.8: Managed Devices SHOULD be required to register MFA.
```

They assume you're running a Hybrid environment and that you tick both **[v] Require device to be marked as compliant** and **[v] Require Microsoft Entra hybrid joined device**. Many of our customers are moving away from Hybrid or at least moving their devices to Entra ID Joined so the **[v] Require Microsoft Entra hybrid joined device** is no longer required for them.

Maybe change the test to just require one of them if that's suitable for your environment?

Snozzberries:
It is a thought I fought with a fair bit while transcribing the original rego to pester tests. If we deviate our tests too far from the CISA controls it causes an issue and as we aren't authoritative for them we don't want to mislead any users of Maester.

For [MS.AAD.3.7](https://github.com/maester365/maester/blob/main/powershell/public/CISA/Entra/Test-MtCisaManagedDevice.ps1#L28) & [MS.AAD.3.8](https://github.com/maester365/maester/blob/main/powershell/public/CISA/Entra/Test-MtCisaManagedDeviceRegistration.ps1#L28) The logic should test for `OR` on the CA policy config, are you seeing a different outcome?

JonesMikael:
Thanks for clarifying. I totally agree that we should follow good official regulations/practices - makes our life so much easier when describing our setup. [CISA Strong Authentication & Secure Registration - MS.AAD.3.7v1](https://github.com/cisagov/ScubaGear/blob/main/PowerShell/ScubaGear/baselines/aad.md#msaad37v1) states this so it's rather them who should change it :)

What I see in most deployments are customers in hybrid environments are moving to Entra ID Joined + Compliance policies in place to make sure the device is Compliant and that is what you want to require. When this in place, most customers leave all Group Policies behind or even delete them and simply don't care about on-prem joined devices. if you leave **[v] Require Microsoft Entra hybrid joined device** you risk that a perpetrator could join a rouge device to your on-prem Active Directory and be able to access your cloud environment through that.

Of course, I could simply change the test to suit our environment.

[...]

PR #200, opened 2024-05-29
Fix#194

Snozzberries:
This is to add the bypass for a compliant device exclusively rather than requiring compliant device OR hybrid joined referenced in #194.

There may be a better way to handle the Pester logic, but this seemed to work as an initial pass.


merill (approved):
This is a good way to handle the custom option. Thanks @Snozzberries!

Issue #198, opened 2024-05-28
Documentation Request - CISA Tests


Reece-McDowell:
I've been writing a write-up on this tooling internally. I'm a big fan of this framework so far and looking to roll it out! However, I noticed that the CISA tests are not mentioned on Github or the website. Could some information about the source of the tests, much like the other two test folders, be added?

Snozzberries:
Thanks for this @Reece-McDowell, a mapping of the test to control ID has been added for initial documentation. If you have additional suggestions, let us know.

PR #201, opened 2024-05-29
A first pass at CISA docs

Snozzberries:
This will give at least a reference of cmdlet to CISA control and let us build on in the future.

The init.ps1 file was just my short hand to create the mapping quickly, we can remove that file if it interferes with docusaurus. 

Ref #198


merill:
Nice work on the docs Michael. Thanks! 

Yes, we might need to move the init.ps1 from the docs folder. 

The ideal place to put this would be in the /build folder. I propose making a /build/cisa folder like we have for /build/eidsca 

For the sidebar label we can drop the **CISA** since it will be in the category node before it.

Merill 2024-06-01 at 20 34 00@2x

We will also need to add this to sidebars.js to create the left hand node.

```
        {
          type: "category",
          label: "🦅 CISA",
          link: {type: 'doc', id: 'tests/cisa/overview'},
          collapsed: true,
          collapsible: true,
          items: [{ type: "autogenerated", dirName: "tests/cisa" }],
        },
```

merill (changes requested):
Nice work on the docs Michael. Thanks! 

Yes, we might need to move the init.ps1 from the docs folder. 

The ideal place to put this would be in the /build folder. I propose making a /build/cisa folder like we have for /build/eidsca 

For the sidebar label we can drop the **CISA** since it will be in the category node before it.

Merill 2024-06-01 at 20 34 00@2x

We will also need to add this to sidebars.js to create the left hand node.

```
        {
          type: "category",
          label: "🦅 CISA",
          link: {type: 'doc', id: 'tests/cisa/overview'},
          collapsed: true,
          collapsible: true,
          items: [{ type: "autogenerated", dirName: "tests/cisa" }],
        },
```

Snozzberries:
Awesome, thanks for the feedback @merill I added a commit.

merill (approved):
Perfect. Thanks @Snozzberries !

Issue #118, opened 2024-04-16
EIDSCA.PR05: Default Settings - Password Rule Settings - Smart Lockout - Lockout duration in seconds - Not correct if using PTA


Azdamus:
The test reports as FAILED if value is above 60 but if your environment uses Passthrough Authentication - then the value needs to be above the on-prem value. 
_“When using pass-through authentication, the following considerations apply: The Microsoft Entra lockout
duration must be set longer than the AD DS account lockout duration. The Microsoft Entra duration is set in
seconds, while the AD DS duration is set in minutes.”_

Cloud-Architekt:
Good catch! I will add some conditions to skip the recommended value if PTA is configured. Therefore, I need to add a logic to identify environments with configured PTA. I start working on this one very soon...

jarboed:
Isn't "Should -BeGreaterOrEqual 60" a more appropriate test here than "Should -Be 60" anyway, regardless of PTA?

I'm set to a lockout duration of 90 seconds, for example, and failing the test because it's not exactly 60 seconds.

Cloud-Architekt:
Thanks for the feedback, @jarboed . I will check to add support for additional ShouldOperator to the template, so that we can implement BeGreaterOrEqual.

Cloud-Architekt:
The issue will be fixed in the next version by adding support for greater or equal values which cover lockout duration value higher than 60 seconds.

PR #205, opened 2024-05-31
Support for additional should operators in EIDSCA tests

Cloud-Architekt:
For some checks, it is not an option to compare a unique value (`Should -be`). Therefore, I've added support for other should operators (e.g., BeGreaterThan). This change requires to change some parts of the code around comparison to allow different operators (e.g., PowerShell operators with hard-coded `-eq` or generated description in the text).

This enhancement will resolve the following issue:
https://github.com/maester365/maester/issues/118


merill (approved):
Looks good @Cloud-Architekt!

PR #196, opened 2024-05-28
#184, Added Identity switch; Added ParameterSets; Configured default …

Snozzberries:
…for Graph to make non-breaking; Added Azure

This provides a framework for beginning to add additional APIs as part of the connection process. Ref option 2 in #184. For ref option 1, I will build off this.

Issue #184, opened 2024-05-20
DISCUSSION: Path to enable additional APIs for tests beyond Graph API


Snozzberries:
I would like to open a discussion around expanding Maester's capabilities to include tests for M365 beyond the Graph API (e.g.,  Exchange Online, SharePoint Online, Defender, Azure, etc...).

Here are a few initial working options:
1. Add tests, but skip unless the API in use is authenticated and available. (i.e., Manual out of band authentication)
2. Add switches to `Connect-Maester` to support proxying additional modules like `Az.Accounts` similar to how it handles Graph today.
3. Utilize an App Registration for these additional APIs, leverage the Az module to create tokens for alternative APIs based on the `.default` authorizations of the app. Any API not supported by Az will be a limitation.
4. Transition to an alternative MSAL manager, such as MSAL.PS or .Net implementation.

Please comment with additional options, any strong preference and logic for a specific option, or additional dependencies or considerations as any option is implemented.

merill:
Whatever we decide we do need to include more details in the docs on writing tests that are not Microsoft Graph and show samples of #1 can be done today (e.g. Connect-AzAccount and calling tests based on Az)

dbrinkmann123:
Thanks @Snozzberries for the comment. Agree wholeheartedly with this as we'd like to use a single platform for testing / assessing environments continuously against a set of best practices. I'd like to see more of SCuBA tests for other workloads folded in as well as other applicable ones. As a MSP we desire to host this infrastructure and run this against our customer tenants.

PR #206, opened 2024-06-01
Multiconnect Added to Connect-Maester

Snozzberries:
Fix to #196 to address providing a framework for beginning to add additional APIs as part of the connection process. Ref option 2 in #184. For ref option 1, I will build off this.

### DIFF
--- a/powershell/public/Connect-Maester.ps1
+++ b/powershell/public/Connect-Maester.ps1
@@ -16,7 +16,7 @@
 
 Function Connect-Maester {
    [Alias("Connect-MtGraph", "Connect-MtMaester")]
-   [CmdletBinding()]
+   [CmdletBinding(DefaultParameterSetName="Graph")]
    param(
       # If specified, the cmdlet will include the scope to send email (Mail.Send).
       [switch] $SendMail,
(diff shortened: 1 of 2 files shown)


powershell/public/Connect-Maester.ps1
```diff
@@ -16,19 +16,54 @@
...
+      $_.Value -eq $true
+   })
+   if($Graph -or $default){
+      Write-Verbose "Connecting to Microsoft Graph"
+      Connect-MgGraph -Scopes (Get-MtGraphScope -SendMail:$SendMail) -NoWelcome -UseDeviceCode:$UseDeviceCode -Environment $Environment
+   }
+
+   if($Azure){
+      Write-Verbose "Connecting to Microsoft Azure"
+      Connect-AzAccount -SkipContextPopulation -UseDesciveAuthentication:$UseDeviceCode -Environment $AzureEnvironment
```
merill: Typo? UseDeviceAuthentication

merill (changes requested):
This is a good start @Snozzberries 

Some comments:

Typo in UseDe**scive**Authentication

Instead of a flag for each service what do you think of an array of services. This will let us expand to include support for other modules (Exchange, Teams, etc..)

Something like
Connect-Maester -Service Graph,Azure,Exchange
Connect-Maester -Service All
etc..

In Disconnect-Maester we might want to check if the connection was made to a particular service before calling it's disconnect method. 

I'm thinking we should store the -Service parameter in a module variable and call disconnect on the ones that Connect-Maester connected to.

We don't want to break an existing Az connection if they are not using it with Maester. 

BTW It will be good to include at least a few examples for the new parameter to make it easy to discover in the docs.

Also the .mdx doesn't need to be manually edited, it will be auto generated by the build.

Snozzberries:
> Instead of a flag for each service what do you think of an array of services. This will let us expand to include support for other modules (Exchange, Teams, etc..)

I am 50/50 on this one. I like the switch approach as it lets us set distinct parameter relationships, like `$AzureEnvironment` vs `$Environment`, but it makes it really bulky. I think the array could provide a simpler experience for most scenarios though, and things like environments, device code, etc... may just need an abstraction layer that references the approved cmdlet parameters.

Will update and push shortly. All great feedback, thank you.

merill (approved):
This looks good. Thanks @Snozzberries!

PR #210, opened 2024-06-03
Update azure-devops.md

iranogajec:
added the yaml for VMSS based DevOps self-hosted agents.


merill (approved):
Love the expanded support for VMSS agents. Thank you so much for taking the time to contribute this @iranogajec

FYI I added VMSS as a tab for better discoverability and to keep the length of the page short.

PR #214, opened 2024-06-05
Add 4.1; First Azure test

Snozzberries:
In ref #169 adding in Azure Diagnostic Settings validation for log forwarding CISA control 4.1.


merill (changes requested):
This is a good start @Snozzberries 

I couldn't get it working however.

Currently fails with: 
```
❯ Test-MtCisaDiagnosticSettings                                   
OperationStopped: /Users/merill/GitHub/maester/powershell/public/CISA/Entra/Test-MtCisaDiagnosticSettings.ps1:52
Line |
  52 |      foreach($log in $logs.Keys){
     |              ~~~~
     | Collection was modified; enumeration operation may not execute.
```

In the CISA tests page it might be a good idea to include a note on how to connect to Az, Exchange and other modules as we expand the scope.

Snozzberries:
Thanks for the call out here @merill. I reworked the function and tested in Pwsh and Windows PowerShell. Hoping it is more consistent now.

I added some additional content to the docs too. I wasn't positive how MDX would handle nested code blocks so I added it just as fenced text instead.

merill (approved):
LGTM

Issue #60, opened 2024-04-10
Corrections on findings


ThoughtContagion:
Not sure if this in intentional or not, but I've found several instances of failed tests where the configured setting is the recommended setting. I will gather the list of these findings and place them in a comment when I finish reviewing the generated report.
![image](https://github.com/maester365/maester/assets/88730003/1e40d7a7-2dd4-4205-b102-06ecf7e9ab42)

There are also some findings that return a value of `****` where the setting does not exist as its queried in the tenant
![image](https://github.com/maester365/maester/assets/88730003/28cd1006-f6a4-4e7f-a03b-7d7f2a48fb1e)
![image](https://github.com/maester365/maester/assets/88730003/3a95d567-ce40-4f2f-8cdf-f498982948c9)

I'm also unsure what some findings are looking for as they trigger a failed result, but the policies/configurations are there.
For example
![image](https://github.com/maester365/maester/assets/88730003/36fc3051-cef4-4f56-a522-d61766d1eec6)
![image](https://github.com/maester365/maester/assets/88730003/2b2699b6-f127-4ac8-8609-da226cb9db1b)

This looks like a fantastic tool, thank you for all your hard work!

ThoughtContagion:
Additional findings that are marked 'Failed' when the recommended setting is configured:
EIDSCA.AT02: Authentication Method - Temporary Access Pass - One-time.

f-bader:
Thank you @ThoughtContagion we will take a closer look

jordan-m-hoff:
+1
I have experienced these same findings. I ran connected with a Global Admin account.

dnsinit:
Just wanted to fill in with some more findings @f-bader : 
EIDSCA.PR02,03,05,06 show as failed if they haven't been touched (meaning they're undefined/null in Graph, default values are also what's recommended).
![image](https://github.com/maester365/maester/assets/46563545/ce34af3e-20c3-4b26-9804-e6ac42e9386f)

Same applies to EIDSCA.ST08 & EIDSCA.ST09.

Below probably needs to be $testResult = $tenantValue -eq 'True' -or '' since empty string being returned should mean default behaviour (True?) aswell.
![image](https://github.com/maester365/maester/assets/46563545/37883b04-9bfb-4b59-bec2-fdebf2c13659)

f-bader:
[...]

PR #220, opened 2024-06-07
Added logic to identify if tenant value is null and seems to be default

Cloud-Architekt:
Added logic to identify "null" values of tenant settings which exists if Microsoft default setting value will be used.
Test result will show recommendation to set the configuration explicitly which allows to compare and validate configuration independently what Microsoft has been set as default.

https://github.com/maester365/maester/issues/60


merill (approved):
Looks good to me!

PR #219, opened 2024-06-07
First Exchange Online Test

Snozzberries:
- Added `Connect-ExchangeOnline` support to `Connect-Maester`
- Added MS.EXO.1.1v1
- Updated template files

Builds on #184 and #206

PR #226, opened 2024-06-16
renamed entra docs and added exo docs, content of entra docs matches …

Snozzberries:
…#214 but may need some tweaks

Building on #219.

Ref #214, the file rename of readme.md to entra.md will cause a merge conflict.


merill (approved):
Thanks a lot @Snozzberries!

I moved the important notes to the main Installation guide page and added pointers from the CISA pages to the install guide. 

A neat enhancement I can think of, is in Invoke-Maester if someone runs the CISA tag then we detect if the optional modules are not installed then link them to the docs page.

Even better enhancement would be to add an Install-OptionalModules and run that to install all the optional modules without the user having to copy and paste commands from the docs.

Issue #195, opened 2024-05-28
MS.AAD.7.3. Declares failed test on environment that has been cloud-only since initial creation


dbenn3268:
I have a tenant that was created in (I believe) 2017 but I know for a fact has never been put into a hybrid configuration or had any hybrid users. By MS.AAD.7.3. failing does that intend to imply I have a hybrid user with a permanent role assigned? If so, how do I get that action to pass if I do not have hybrid users?

Installed Maester on fresh VM instance before running this resultant test.

![image](https://github.com/maester365/maester/assets/106715637/8ef9438c-6be2-40c3-af08-d1318661cb0d)

![image](https://github.com/maester365/maester/assets/106715637/80d77db3-b8e8-4498-afb1-413cd6f9e9d3)

Snozzberries:
Here is the code block the test is calling. Can you run it interactively and confirm that `$users` doesn't contain any object with `onPremisesSyncEnabled`. There shouldn't be a discrepancy, but it may help to confirm.

Ref https://github.com/maester365/maester/blob/main/powershell/public/CISA/Entra/Test-MtCisaCloudGlobalAdmin.ps1

```
    $role = Get-MtRole | Where-Object {`
        $_.id -eq "62e90394-69f5-4237-9190-012177145e10" }

    $assignments = Get-MtRoleMember -roleId $role.id -All

    $globalAdministrators = $assignments | Where-Object {`
        $_.'@odata.type' -eq "#microsoft.graph.user"
    }

    $body = @{
        requests = @()
    }
    $requests = @()
    $i = 0
    foreach($admin in $globalAdministrators){
        $request = @{
            id     = "$i"
            method = "GET"
            url    = "/users/$($admin.id)?`$select=id,displayName,onPremisesSyncEnabled"
        }
        $requests += $request
        $i++
    }
    $body.requests = $requests

    $usersSplat = @{
        Method     = "POST"
        Body       = $($body|ConvertTo-Json)
        Uri        = "https://graph.microsoft.com/v1.0/`$batch"
        OutputType = "PSObject"
    }
    $users = (Invoke-MgGraphRequest @usersSplat).responses.body

    $result = $users | Where-Object {`
        $_.onPremisesSyncEnabled -eq $true
    }

    $testResult = $result.Count -eq 0
```

Snozzberries:
Interesting, it seems like it is an issue with `Get-MtRoleMember` permissions.

[...]

PR #227, opened 2024-06-16
Privileged permissions

Snozzberries:
Here is a first pass at a concept of enabling Maester with optional privileged permissions.

The initial use case is to address #195 but future use cases may arise as well.


merill:
@Snozzberries Made a few updates 

- Switched to use the OOTB batch request option in Invoke-MtGraphRequest (it also fixes the max requests in a batch request by creating multiple batches if there are more that 10 requests).
- Excluded expired/revoked states which showed incorrect users
- Switched the link to take directly to the user's role assignment blade.

I think the help/context for this might need to be updated. It refers to running commands to find the users but in our case we are showing the list of users already.

PR #229, opened 2024-06-17
"Add tags for all maester tests"

f-bader:
This pull request adds tags to all maester tests to make it easier to include or exclude them.


f-bader:
@merill please check and approve. Only added tags to comply with the other tests and make it easy to exclude certain built-in tests

merill (approved):
Nice. This is great.

PR #231, opened 2024-06-18
Update Test-AppManagementPolicies.Tests.ps1

gerlasch:
added license check for entra id governance licensing, skipping test if license not available


f-bader:
@Snozzberries quick question about the governance return code. Isn't this a different SKU from Entra ID P1 or P2. If so, we should add it as a separate product and not as part of Entra ID, or support a multi array return values. What do you think?

https://github.com/maester365/maester/blob/9e0bd8e974eaa8a822ae4b1f32cf84db9012bf6a/powershell/public/Get-MtLicenseInformation.ps1#L30

merill (approved):
I have fixed this one to check the Workload ID SKU but yes, the Governance one will need to be updated to have EntraIDGovernance as another parameter fyi @Snozzberries

PR #254, opened 2024-06-19
Update azure-automation.md

DanielBradley1:
Corrected issues with text


DanielBradley1 (commented):
Change connection string to -Identity

PR #259, opened 2024-06-20
Fix mt.1027 bug and fixed test description output from Test-MtPrivPermanentDirectoryRole if it passed

tdcthosc:


f-bader:
Hi @tdcthosc could you explain the reasoning for disabling the cache

tdcthosc:
> Hi @tdcthosc could you explain the reasoning for disabling the cache

Hi @f-bader I agree that we need to utilize the cache. I will remove the disable cache switches.

Issue #255, opened 2024-06-19
Maester appears to be bundling an unofficial  version of Pester 5.6


timrefw:
Install-Module -Name Maester

Untrusted repository
You are installing the modules from an untrusted repository. If you trust this repository, change its
InstallationPolicy value by running the Set-PSRepository cmdlet. Are you sure you want to install the modules from
'PSGallery'?
[Y] Yes  [A] Yes to All  [N] No  [L] No to All  [S] Suspend  [?] Help (default is "N"): Y

**WARNING: Module 'Pester' version '3.4.0' published by 'CN=Microsoft Windows, O=Microsoft Corporation, L=Redmond, S=Washington, C=US' will be superceded by version '5.6.0' published by 'CN=Jakub Jareš, O=Jakub Jareš, L=Praha, C=CZ'. If you do not trust the new publisher, uninstall the module.**

merill:
This is a known issue on Windows and is because Windows out of the box includes an older version of Pester that was signed by Microsoft. Pester is now open sourced and is maintained by the community. The older version is no longer used.

See the guide at 
https://maester.dev/docs/installation

To avoid the warning run this command before installing Maester.
```powershell
Install-Module Pester -SkipPublisherCheck -Force -Scope CurrentUser
```

SamErde:
@merill, would it be helpful to add an optional `-Prerequisites` switch to `Install-MaesterTests` that includes something like this?

```powershell
if ( ((Get-Module Pester -ListAvailable).Version | Sort-Object -Descending -Top 1) -lt '5.6.0') {
    Install-Module Pester -SkipPublisherCheck -Force -Scope CurrentUser
}
```

merill:
That's a great idea @SamErde  why didn't I think of that!!

Would you like to submit a PR?

SamErde:
Will do!

SamErde:
Why stop at just Pester?! This just became good inspiration for a function to install any missing required module versions.

merill:
Thanks @SamErde the latest version now does the auto-install of Pester during Install-MaesterTests.

PR #265, opened 2024-06-22
Automatically install Pester if needed unless using -SkipPesterCheck

SamErde:
Addresses #255 and adds enhancement by adding `-Prerequisites` parameter to `Install-MaesterTests`. Checks required modules and installs them if the minimum required version is not already installed.


merill:
Thanks for the PR @SamErde.

A small catch.

The Maester.psd1 file declares Pester as a required module with the specific version.

This means `Install-Module Maester` itself will fail if Pester doesn't get updated to 5.5.0. As a result the user who hits this issue will never get a chance to run the cmdlet/param you defined.

My thinking was to 
- Update the RequiredModule in Pester to not include a specific version. This way MacOS and other environments will install the latest version and Windows will skip trying to auto update and failing
- Make Install-MaesterTests **automatically** check for Pester 5.5.0 and install+import if it's not there.
- Add a -SkipPesterCheck parameter to opt out of the Pester check (instead of -Prerequisite as an opt-in)

https://github.com/maester365/maester/blob/aa79b04cbe012bfc1488eaae30848bdfc251749b/powershell/Maester.psd1#L54

Thoughts?

SamErde:
Yes, that makes perfect sense. That flow will reduce friction for people who are just getting started. I can restructure this PR to reflect that concept.

SamErde:
Ready for review, @merill. Is `Import-Module 'Pester'` needed? I may have missed that, but didn't see it being used previously.

SamErde:
This build passes, but I want to look back at the previous [build validation step](https://github.com/maester365/maester/actions/runs/9636640609/job/26574838743#step:6:16) that failed when I left Pester listed as a required module but removed the version requirement. That change failed (ironically) a Pester test of the module definition file with:

```PowerShell
Error: "The 'ModuleVersion', 'MaximumVersion' and 'RequiredVersion' members do not exist in the hashtable that describes this module. One of these three members must exist, and be assigned a version number in the format 'n.n.n.n'."
```

What’s the best way to skip this specific test so we can leave it listed as a RequiredModule with no version specified?

Alternatively, we could set the minimum version here to be 3.4.0, and then re-check with Install-MaesterTests, but that also has downsides.

SamErde:
@merill, I have done some more testing and have two options that pass Test-ModuleManifest.

- `@{ModuleName = 'Pester'; GUID = 'a699dea5-2c73-4616-a270-1f7abb777e71'; ModuleVersion = '0.0.0'; }`
- `@{ModuleName = 'Pester'; GUID = 'a699dea5-2c73-4616-a270-1f7abb777e71'; ModuleVersion = '3.4.0'; }`

The first passes manifest checks and effectively accepts any version of Pester. This allows `Install-MaesterTests` to install or update it, if necessary. 

The second also passes and looks for the version we expect to find installed out of the box in Windows.

Given the cross-platform use of Maester, do you have a preference for what goes here instead of my last commit that removed the module requirement (I didn't like that).

merill:
Awesome. Thanks for this. I was testing the previous PR and was even testing some workarounds that involved having two .psd1 files with the first one loading the second .psd1 but it was too hacky and not reliable.

I like what you are doing here. I think it doesn't make a difference between the two options. Let's add a comment to note why we set to that version number.

SamErde:
Glad you tested the dual PSD1 approach. I saw one blog post about doing that and it seemed very interesting, but also short on details.

Version "0.0.0" feels more OS agnostic and less misleading than declaring "3.4.0" as the required version. I don't want to trigger anyone on the "it's not actually a version" principle, but it's the direction I lean towards.

SamErde:
Ready to roll. Thanks for the feedback!

merill (approved):
Nice! Thanks @SamErde !!

PR #266, opened 2024-06-22
Updated MD and added comment for Global Admin refs

Snozzberries:
Some additional cleanup from #227 comment.


merill (approved):
Thanks for the updates @Snozzberries

PR #267, opened 2024-06-22
Additional EXO Tests

Snozzberries:
These are all the EXO controls that require only ExchangeOnlineManagement cmdlets. The remaining controls will require additional cmdlets.


merill:
I noticed a bunch of links have the portal deeplinks have tid parameter which takes the user to a different tenant's sign in page. I think the link works without the tid. 

Also if the result has only one item it might be better to skip the tick emoji since the UX already shows an icon. 

Might be cleaner with something like this 

```powershell
    if ($testResult) {
        $testResultMarkdown = "Well done. Your tenant does not have any [Anti-spam IP allow lists]($portalLink)."
    } else {
        $testResultMarkdown = "Your tenant has [Anti-spam IP allow lists]($portalLink).`n`n%TestResult%"
        $policy.IPAllowList | ForEach-Object {`
                $result += "* $_`n"
        }
    }

```

merill (approved):

Snozzberries:
Thanks Merrill!

PR #269, opened 2024-06-23
Added SkipReason support

merill:
Added support to show reason for skipping in report and updated Exchange tests and one Maester test to use this.


powershell/internal/Get-MtSkippedReason.ps1
soulemike, line 1: This looks good, an initial thought was would it be more maintainable as a hash table?

powershell/public/CISA/exchange/Test-MtCisaAntiSpamAllowList.ps1
soulemike, line 1: Does the OutputType need null added? It didn't look like Pester barked at it on these, but I have seen it on others.

tests/Maester/Entra/Test-AppManagementPolicies.Tests.ps1
soulemike, line 1: Was dropping the L-WI tag intentional? I am assuming yes, as it maybe referenced "License-WorkloadIdentity".

merill: My initial plan was to add tags to teh test and have the tests skipped based on the tags. We have a much more flexible way no and having the license check in the cmdlet that does the check is a better outcome.

powershell/internal/Get-MtSkippedReason.ps1
merill: Yeah I want to make the validation dynamic as well based on a hash table.

I'm planning on adding a C# module as part of adding support for export to Excel/C# and was thinking of moving some of code like this to c# so it will be a lot more performant.

PR #271, opened 2024-06-25
Mail Authentication Functions to support CISA EXO SPF/DKIM/DMARC Checks

Snozzberries:
These functions provide a fairly robust way to capture and convert most of the context associated with mail authentication protocols (e.g., DKIM/DMARC/SPF).


michaelmsonne:
Nice add @Snozzberries - then we soon just need to see whats the status for deployments of DANE support in EXO at some point :)

merill (approved):

PR #272, opened 2024-06-25
fix: bump checkout to v4

Arcath:
Minor change, the GitHub Action (which is great btw) uses actions/checkout@v3 which uses Node 16, bumping to v4 uses Node 20 which fixes this deprecation.


michaelmsonne:
Good catch @Arcath - @f-bader 👌

merill (approved):

Issue #228, opened 2024-06-17
Feature Request: Add generated with module version to HTML dashboard


ClassyGull:
Would be useful to see the Maester module version that generated the Maester test results on the html dashboard. Perhaps top right under the time generated component. 

I understand there is a separate process to run for updating tests when using ADO pipeline, but when running from Azure Automation runbook the tests are based on the module version (as far as i can tell) and installed each time. So maybe this would be a param on Invoke-Maester.

The goal is to understand if the latest tests are being run. If possible, displaying the latest version of Maester would also be nice (with hyperlink to module download).

Thanks.

merill:
This is a good feature request. Thanks @ClassyGull

I'll add this to the backlog.

merill:
Many thanks to @GuidoBaijense we now show the version number in the reports and emails.

PR #273, opened 2024-06-25
#228: add module version to html, md and mail reports

GuidoBaijense:


merill (approved):
Thanks @GuidoBaijense this a really good PR!

Love that you were able to follow the guide and update the mail template.

On the Report did you forget to check in the app.jsx file? The instructions to update the html report is in the /report/readme.md file

Either way I tweaked the design and moved the version number to the footer and removed the label so it just shows the version number.

GuidoBaijense:
@merill I didn't forget to check it in. I just did not understand yet how to work with these kind of templates and edited the wrong file. Was wondering why it rolled back after a build... Readme does not mention the app.jsx file, however it probably should be common knowledge for someone that works with these kinds of frameworks.

#228 mentioned a link to the newest module. Do you think it's better to just show the version number?

PR #279, opened 2024-06-27
Fix handling of `Invoke-Maester` `Output*` params

### LINKED ISSUE #278: Output* arguments on `Invoke-Maester` being clobbered
### Issue
I encountered this issue with v0.1. I'm running in WSL, but I don't think there is anything platform specific about this issue. The issue is that the `OutputJsonFile` argument is ignored when running `Invoke-Maester`. I suspect this is true for `OutputHtmlFile` and `OutputMarkdownFile` but I have only confirmed the case using `OutputJsonFile`.

### Steps to reproduce
To reproduce, install Maester, configure credentials and then run the following:
```pwsh
Invoke-Maester -NonInteractive -OutputJsonFile "test-result.json"
```
The `OutputJsonFile` argument is **not** honored, and it reverts to generating test results using the default strategy, i.e., the output is as follows:
```bash
test-results
├── TestResults-2024-06-26-220909.html
├── TestResults-2024-06-26-220909.json
├── TestResults-2024-06-26-220909.md
```
### Expected result
The docs for `OutputFolder` state:
> The folder to save the test results. If PassThru and no -Output* is set, defaults to ./test-results. If set, other -Output* parameters are ignored and all formats will be generated (markdown, html, json) with a timestamp and saved in the folder. 

The docs for `OutputJsonFile` state:
> The path to the file to save the test results in json format. The filename should include a .json extension.

Given that I did **not** specify the `OutputFolder` argument and **did** specify the `OutputJsonFile` argument, the expected output of running the same command should have been:
```bash
<pwd>
├── test-result.json
```

pkothare:
This pull request fixes the handling of `Output*` parameters in the `Invoke-Maester` function. There were 2 issues identified:

1. The `$out.OutputHtmlFile` property, was incorrectly referenced in `ValidateAndSetOutputFiles`. Prior to this change, the script referenced an invalid member `$out.OutputFile`. It was updated to the correct member reference of `$out.OutputHtmlFile` 

2. The conditional logic in `ValidateAndSetOutputFiles` now checks that _if and only if_ all the following are true:

   1. `OutputFolder` is null or empty
   2. `OutputJsonFile`, `OutputHtmlFile` and `OutputMarkdownFile` are all null or empty

   then the value of `OutputFolder` will be set to `./test-results`, in accordance with the docs: https://maester.dev/docs/commands/Invoke-Maester#-outputfolder. Checking `$PassThru` is inconsequential and was removed from the conditional logic because either way, files will be output to the file system.
   
Resolves #278

### DIFF
--- a/powershell/public/Invoke-Maester.ps1
+++ b/powershell/public/Invoke-Maester.ps1
@@ -142,7 +142,7 @@ Function Invoke-Maester {
     function ValidateAndSetOutputFiles($out) {
         $result = $null
         if (![string]::IsNullOrEmpty($out.OutputHtmlFile)) {
-            if ($out.OutputFile.EndsWith(".html") -eq $false) {
+            if ($out.OutputHtmlFile.EndsWith(".html") -eq $false) {
                 $result = "The OutputHtmlFile parameter must have an .html extension."
             }
         }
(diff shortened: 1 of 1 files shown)


pkothare:
IMO this sort of scenario handling should really be constrained using [parameter sets](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_parameter_sets?view=powershell-7.4), but I don't want to go against the established approach in the code. 

Using parameter sets also requires docs to be updated and further design review that should be more carefully considered. Let me know if that is desired with Maester, and I can open a new PR with a comprehensive set of changes. This PR directly addresses the issue at hand, and I hope that my interpretation of the intent is correct here.

merill (approved):
Thanks for the fix @pkothare much appreciated.

As you suggested, it would be good to use ParameterSets. A PR that converts to parameter sets but keeps the existing interface would be really good.

PR #280, opened 2024-06-27
Enhance mt.1024 entra recommendations

### LINKED ISSUE #189: MT.1024 - Incorporate results of Entra Recommendation in output
We have been working through this Maester report on an ongoing basis. One thing that we have noticed as an issue is that although it gives a more verbose output in some cases for MT.1024 Entra recommendations than the portal sometimes it gives a verbose count of affected users / resources but doesn't output those values. 
An example of this would be:

MT.1024: Entra Recommendation - Require multifactor authentication for administrative roles.
You have 6 of 68 users with administrative roles that aren’t registered and protected with MFA.

How can we find out who these users are so we can see why the existing CA policies aren't covering them? The Entra recommendation in the portal is less helpful than the report because it doesn't even tell the number of affected users.

tdcthosc:
Partly fixes #189 where Entra Recommendations include impacted resources, however it would seem that user related recommendations does not include users.

This PR includes action URLs into the remediation steps. Some steps ended abruptly because the URL was part of the sentence. It also enhances many other remediation steps with relevant URLs.


merill (approved):
This is amazing! Love it.

PR #298, opened 2024-07-03
Added Link to Docs for how to Fix from Entra Portal

Appelcloud:
I have added a row in the "how to fix" part, for how the fix can be applied from Entra portal.

What do you think about this? :-)


merill:
Thanks for the PR @Appelcloud 

Would you be able to submit this PR to the EIDSCA project? The files you edited here are auto-generated and will be replaced at the next run.

See the guide here
https://maester.dev/docs/contributing#updating-eidsca-tests-and-documentation

The file you need to edit and submit the PR is the json file below. Once you update there we will regenerate the command and the links will flow through. 

@Cloud-Architekt fyi

https://github.com/Cloud-Architekt/AzureAD-Attack-Defense/blob/AADSCAv4/config/EidscaConfig.json

merill (changes requested):
Please see comment above.

Cloud-Architekt:
@Appelcloud : This is a great idea and I've created a PR to support "how to fix" links from the EidscaConfig.json file.
A new property (`HowToFix`) is available in the schema and I have already added your links to the file.
Would you please check in the following PR and share your thoughts in the other PR if this change would covers your request. It would be great if you could help to add more "how to fix" links to the various checks. Thanks again!

PR #300, opened 2024-07-04
#5: Send report as Teams channel message

GuidoBaijense:
Had some fun and learned something new. This PR adds functionality to send Microsoft Teams Channel Messages with the Maester-report summary as described in issue #5 . I used Send-MtMail.ps1 as reference and tried to update all documentation. 

>Note: Following MS Teams Terms of Use, it is not allowed to use Teams as a logging system. Only messages that will be read are allowed. Also, there is no official support for sending teams messages with application permissions. However, this might be possible with the migration permissions. 

Next might be a switch to send the message to a user instead of a teams channel. I think that it's better to spend my time on something more useful :).

Since I didn't want to create a new release, I haven't updated the Maester.psd1 with new guids or version numbers.


powershell/public/Invoke-Maester.ps1
```diff
@@ -129,6 +129,14 @@ Function Invoke-Maester {
         # This is required when using application permissions.
         [string] $MailUserId,
 
+        # Optional. The Team Id of where the Teams Channel Message should be send to. e.g. 5ba5cbbb-675e-4a8d-a382-a52960a18b4d
+        # No message will be sent if this parameter is not provided.
+        [string[]] $TeamId,
```
merill: Does this need to be a string array?

powershell/public/Invoke-Maester.ps1
```diff
@@ -129,6 +129,14 @@ Function Invoke-Maester {
...
+
+        # Optional. The Channel Id of where the Teams Channel Message should be send to. e.g. 19:4a95f7d8db4c4e7fae857bcebe0623e6@thread.tacv2
+        # No message will be sent if this parameter is not provided.
+        [string[]] $ChannelId,
```
merill: Same here for string array. Shouldn't it be a [string] like the other parameters?

merill: Let's call this TeamChanelId so we know it's related to the other Teams property.

powershell/public/Get-MtGraphScope.ps1
```diff
@@ -28,6 +33,9 @@ Function Get-MtGraphScope {
         # If specified, the cmdlet will include the scope to send email (Mail.Send).
         [Parameter(Mandatory = $false)]
         [switch] $SendMail,
+        # If specified, the cmdlet will include the scope to send Teams Channel Messages (ChannelMessage.Send).
+        [Parameter(Mandatory = $false)]
+        [switch] $SendChannelMessage,
```
merill: Let's call this SendTeamsMessage to make it clear. Not many folks would understand Channel.

powershell/public/Send-MtTeamsChannelMessage.ps1
```diff
@@ -0,0 +1,153 @@
...
+        }
+
+    $notRunCount = $MaesterResults.SkippedCount
+    if ([string]::IsNullOrEmpty($MaesterResults.SkippedCount)) { $notRunCount = "-" }
+
+    $adaptiveCardData = @{
+            title       = $Subject
+            description = "Results for Maester Test run of $($MaesterResults.ExecutedAt)"
+            run         = @{
+                tenantName    = $MaesterResults.TenantName
```
merill: Can we keep the variable name case uniform? Some start with lowercase (e.g. moduleVersion) and others start with upper case (e.g. NotRunCount).

merill (changes requested):
This is really cool. Thanks a lot. I've made a few suggestions. Would you be able to resubmit with the changes?

Tx!

GuidoBaijense:
> This is really cool. Thanks a lot. I've made a few suggestions. Would you be able to resubmit with the changes?
> 
> Tx!

Good review! Thank you. I resolved all of your comments. (as far as I can see)

merill (approved):
Excellent!

PR #305, opened 2024-07-07
Exo security

Snozzberries:
* Support for SCC connection
* First SCC control, 8.1


merill (approved):
Nice addition @Snozzberries.

Issue #292, opened 2024-07-02
EXO DNS Tests do not support cross platform


Snozzberries:
Currently the tests added in #271 do not support cross-platform use given a dependency on `Resolve-DnsName`.

A check for `-not $IsWindows` and then use of a cross-platform library will be necessary, such as https://github.com/rmbolger/DnsClient-PS.

Search Refs:
Test-MtCisaSpfRestriction	MS.EXO.2.1v1
Test-MtCisaSpfDirective	MS.EXO.2.2v1
Test-MtCisaDkim	MS.EXO.3.1v1
Test-MtCisaDmarcRecordExist	MS.EXO.4.1v1
Test-MtCisaDmarcRecordReject	MS.EXO.4.2v1
Test-MtCisaDmarcAggregateCisa	MS.EXO.4.3v1
Test-MtCisaDmarcReport	MS.EXO.4.4v1

Snozzberries:
With #322 this is now supported.

![image](https://github.com/maester365/maester/assets/431932/bdf38beb-8ad5-4385-85d0-11549ae5f283)

PR #308, opened 2024-07-07
Fix to Team* parameters and Service Parameter logic

Snozzberries:
* Team* parameters default values allow for null match
* Service parameter allows for All to not also require ExchangeOnline for SCC to be functional


Snozzberries:
Commit [2683f62](https://github.com/maester365/maester/pull/308/commits/2683f62b45bd48f5da342c3bf32912da990e40bd) fixes #292

PR #310, opened 2024-07-08
Do not exit shell when cancelling Update-MaesterTests

### LINKED ISSUE #160: Update-MaesterTests exits session, preference toward return may be better
https://github.com/maester365/maester/blob/f90399bde1849ddce74490fe40c3d2511b50db0e/powershell/internal/Update-MtMaesterTests.ps1#L42

Would need to do some more testing, but currently the hard exit will kill the processes. A return may allow better error handling and feedback.

SamErde:
Replace `exit` with `return` so the function returns to the prompt instead of closing the shell completely. Resolves #160.


merill (approved):
Good catch. Thanks @SamErde

Issue #173, opened 2024-05-13
Update-MaesterTests removing my folders


dafuni:
I am running the Maester updates and receiving a message that all my folders will be removed. Can I please get a clarification on this?
![image](https://github.com/maester365/maester/assets/5445383/85c471e7-10d1-4d24-904a-2c06a966dad9)

DrPye:
Ok so I looked into this for you, and I'm not sure if its by design or not but...

The issue is that your not providing a path so it uses your current working directory. now even though there is a cmdlet that gets the maester-tests folder path, that path is not used when getting a list of folders to use. It instead uses your current working directory as your not passing a path.

I've not tested this but a quick work around would be to run ```Update-MaesterTests -path "C:\Users\{username}\maester-tests"```

Let me know how you get on!

merill:
Yes please run this command inside the folder that contains your Maester tests or as @DrPye suggested provide a path.

I think we should detect if there is a folder called maester-tests and prompt if the user wishes to update that folder. That way we can avoid this scenario if the folder already has tests.

dafuni:
That worked. Thanks
![image](https://github.com/maester365/maester/assets/5445383/35c64acb-d706-4963-bb46-f895c809c6a0)

merill:
Updating this to be implemented as an enhancement.

Update-MaesterTests should be updated to

- Check if the current folder contains more folders than the four standard folders in our tests folder (ie CISA, Custom, EIDSCA, Maester). 
  - If the four folder are missing or if there are more than four folders found it should error out with a message asking the user to run the command inside the maester-tests folder.

DrPye:
@merill looking at the code we might have a simpler solution already.

If you look at the following snippet from Update-MaesterTests
```pwsh
    $MaesterTestsPath = Get-MtMaesterTestFolderPath

    if (-not (Test-Path -Path $MaesterTestsPath)) {
        Write-Error "Maester tests not found at $MaesterTestsPath"
        return
    }

    $targetFolderExists = (Test-Path -Path $Path)

    $installOrUpdate = if ($Install) { "installed" } else { "updated" }
    if ($targetFolderExists) {
        # Check if the folder already exists and prompt user to confirm overwrite.
        $itemsToDelete = Get-ChildItem -Path $Path -Exclude "Custom"
```

We could just update ```Get-MtMaesterTestFolderPath``` to accept a path param and default to ```psscriptroot``` if not provided.

[...]

PR #312, opened 2024-07-08
Only delete known Maester tests with Update-MaesterTests

SamErde:
Checks the known Maester test names from the installed module path and only deletes those child object names instead of everything in $Path. Provides a resolution for #173.

### DIFF
--- a/powershell/internal/Update-MtMaesterTests.ps1
+++ b/powershell/internal/Update-MtMaesterTests.ps1
@@ -62,7 +62,6 @@ Function Update-MtMaesterTests {
         }
     }
 
-    $MaesterTestsPath = Get-MtMaesterTestFolderPath
     Copy-Item -Path $MaesterTestsPath\* -Destination $Path -Recurse -Force
 
     $message = "Run `Connect-Maester` to sign in and then run `Invoke-Maester` to start testing."
--- a/powershell/internal/Update-MtMaesterTests.ps1
+++ b/powershell/internal/Update-MtMaesterTests.ps1
@@ -23,12 +23,14 @@ Function Update-MtMaesterTests {
         return
     }
 
+    $MaesterTests = (Get-ChildItem -Path $MaesterTestsPath -Exclude 'Custom').Name
+
     $targetFolderExists = (Test-Path -Path $Path -PathType Container)
 
     $installOrUpdate = if ($Install) { "installed" } else { "updated" }
     if ($targetFolderExists) {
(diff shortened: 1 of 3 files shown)


merill:
This is good Sam. Thanks. Do you have thoughts on how to handle scenarios where we delete/retire tests?

SamErde:
Yes, I do have some early thoughts on how to handle removed or retired tests. I was also thinking of adding a test update process that can be separate from the module update process (an -Online parameter for updating tests).

Handling removed tests could be handled in a few ways:
- Maintain an inventory of active tests, either manually as they change or via a script that checks test files/folders during a build step.
- Leave the test files, but append ".Retired" (or similar) to the .PS1 and .MD file extension.
- Add status metadata in each of the test files themselves, with logic in the Invoke and Update functions to skip/remove retired tests.
- Move them to a "deprecated" folder and exclude that folder when invoking the tests.
- Probably other possibilities as well.

I believe any of these ideas could benefit from an optional online update method that doesn't depend on the module being updated, but I'm sure there are factors that need to be considered.

When considering module download size, there will be benefits to simply removing retired tests from new releases.

merill:
I'm thinking we can add a hashtable to update-maester with a list of all .tests.ps1 files that we've removed. Then it can do a targeted delete of just those.

SamErde:
Would you generate the hash table in some automatic fashion by comparing builds or releases, or just add the information by hand when necessary?

SamErde:
> I'm thinking we can add a hashtable to update-maester with a list of all .tests.ps1 files that we've removed. Then it can do a targeted delete of just those.

A thought: using a hash table to selectively remove retired tests will be a change to the current wipe/replace approach. If it works, would you like to test and merge this PR, and then follow-up with a separate PR to rethink how updates/changes to tests can be handled? This PR includes:

- Error handling for directory creation and file copy operations.
- Relative path expansion and validation.
- Enhanced verbose output during the new/copy operations.
- Checks to ensure that the only things that get deleted will be files and folders that match the names present in the module's installed `maester-tests` directory.
- Replaced `exit` with `return` statements.
- Simplification of the function and removal of a redundant line of code.
- Added comment-based help.

merill (approved):
💯% agree. This fixes a lot of the issues and makes the process non-destructive.

Also, I doubt we are going to delete any test folders soon so that can come later when we do get to that point.

SamErde:
OK, thanks!

PR #315, opened 2024-07-09
Make PowershellGet optional

deleugpn:
The version `0.1.0` didn't make use of any code that requires the `PowershellGet` module to be installed. When trying out version `0.2.0` I started facing issues due to the use of `Find-Module` not being available.

The use-case I have is the ability to run Maester standard tests + Maester custom tests (user-defined) where we want to avoid users providing unsafe scripts. The environment is stripped down and `PowershellGet` is not installed.

### DIFF
--- a/powershell/internal/ConvertTo-MtMaesterResults.ps1
+++ b/powershell/internal/ConvertTo-MtMaesterResults.ps1
@@ -40,8 +40,13 @@ function ConvertTo-MtMaesterResult {
     $tenantName = GetTenantName
     $account = $mgContext.Account
 
-    $currentVersion = ((Get-Module -Name Maester).Version | Select-Object -Last 1).ToString()
-    $latestVersion = (Find-Module -Name Maester).Version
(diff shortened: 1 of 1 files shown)


powershell/internal/ConvertTo-MtMaesterResults.ps1
```diff
@@ -40,8 +40,13 @@ function ConvertTo-MtMaesterResult {
     $tenantName = GetTenantName
     $account = $mgContext.Account
 
-    $currentVersion = ((Get-Module -Name Maester).Version | Select-Object -Last 1).ToString()
```
SamErde, line 43: Just a thought: you could leave $currentVersion outside of the check for `Find-Module` and still always show it because `Get-Module` is part of Microsoft.PowerShell.Core.

deleugpn: Thanks, implemented!

PR #317, opened 2024-07-11
Fix to property reference on conversion functions

Snozzberries:
Had a bug in the non Windows conversion functions.


merill (approved):
Tx!

PR #318, opened 2024-07-11
Added skip conditions for "settings" APIs in EIDSCA

Cloud-Architekt:
Some of values of the "settings" API in Microsoft Graph have no values in recent created tenants.
Therefore, I've created a skip condition when the related response contains no value for the specific check.


merill (approved):
It would have been nicer if we could do the skip check inside the cmdlet then it would be possible to include the reason for skipping in the final report.

This is good as a first step.

PR #319, opened 2024-07-11
Added support for Should "BeIn" function in Pester in EIDSCA

Cloud-Architekt:
Some checks have more than a single value which can be accepted as "recommended" configuration.
Therefore, I've started to implement support for "BeIn" statements in EIDSCA without a big impact of the existing EIDSCA JSON schema. Defining a value in syntax of a object in the property "RecommendedValue" (string property) will be recognized and seems for me a valid workaround.


Cloud-Architekt:
@merill : Do you like to have a look on this PR? It would be great to have feedback if this could be a valid approach.

merill (approved):
Looks great!

PR #324, opened 2024-07-12
Two bug fixes

Snozzberries:
Updates to get an additional test within scope and to fix the SCC connection logic.

With an accidental easter egg.


merill (approved):
Oops. I missed that one too!

PR #326, opened 2024-07-12
Added support for custom skip reasons in EIDSCA checks

Cloud-Architekt:
- Added support for parameter SkippedCustomReason in `Add-MtTestResultDetail` by @f-bader 
- Added custom text for SkippedCustomReason (also known as SkipReason in EIDSCA JSON
- Updated "How to fix" links

### DIFF
--- a/build/eidsca/Update-EidscaTests.ps1
+++ b/build/eidsca/Update-EidscaTests.ps1
@@ -333,6 +333,25 @@ if ($currentValue -eq '' -or $control.ControlName -eq '') {
     $output = $output -replace '%PSFunctionName%', $psFunctionName
     $output = $output -replace '%PortalDeepLinkMarkdown%', $portalDeepLinkMarkdown
     $output = $output -replace '%GraphDocsUrlMarkdown%', $graphDocsUrlMarkdown
+
+    if ($controlItem.SkipCondition -ne "") {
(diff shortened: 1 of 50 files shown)


Cloud-Architekt:
@merill : As you mentioned in PR https://github.com/maester365/maester/pull/318, the current implementation of skip was the first step. I have worked on an enhanced version to support the already defined SkipReason from the EIDSCA schema.
What do you think about adding the Discovery as variables in scope of the script? This was for me a workaround to use the discovery results in the various checks.

merill:
I'm not quite sure I follow. Did you want the Discovery step to set module level variables?

Cloud-Architekt:
@merill : Yes, as variable on script level. I'm not quite if there's any other solution to use a variable from the discovery section (tests/EIDSCA/Test-EIDSCA.Generated.Tests.ps1) in the various Test-* cmdlets. I would like to avoid moving the same logic to the individual cmdlets. What do you think? Is there any better approach without setting this as script variable?

merill:
If things run in parallel I'm worried we could end up with incorrect runs.

The Pester docs have an example to use -TestCases to pass from Discovery to the test.

https://pester.dev/docs/usage/discovery-and-run#variables-are-not-available-in-test

My thought is to call a cmdlet within the test that can get this info (then it can use the Invoke-MtGraph.. cache) and then skip from within.

Issue #325, opened 2024-07-12
CISA SCuBA v0.2.9 known bugs


Snozzberries:
AAD 8.3, 7.9, 7.8, 7.2, 4.1, 3.3 - Results not properly displaying.

EXO 3.1 - Skipping on Linux, seems similar to DMARC, but needs more testing.

EXO 4.* - `ConvertFrom-MailAuthenticationRecordDmarc` crashes when no answers returned to class init.

Snozzberries:
As of 0.2.20, AAD 7.8 and 7.9 are skipping for EidGov and shouldn't, AAD 7.2 divide by zero description, AAD 4.1 sort by log name.

PR #328, opened 2024-07-13
Fixes for EXO 3, 4 in #325

Snozzberries:
I believe this will address part of #325 but need to test deep functions.


merill (approved):
Thanks Michael!

PR #334, opened 2024-07-13
Fix Teams check in Invoke-Maester

fflaten:
Fixes small bug introduced in #308 that always required teams API permissions. `$null` default value for parameters results in empty string when cast to `[string]` which is never null.

PR reverts default values and checks that both parameters are not null or empty strings before requesting Teams-permissions.

/cc @Snozzberries


Snozzberries:
I think this is a cleaner approach than the default value of $null to force the logic.

Quick cross-platform check:

Windows Core
image

Ubuntu PWSH
image

Windows PowerShell
image

Issue #98, opened 2024-04-14
Suggestions to improve module tests


fflaten:
Great work on this project! While looking through the codebase I had a few suggestions to improve the module's own Pester-tests.

- There's duplicate help tests - **kinda**. Do we need both? Should internal functions be tested?
  - `tests/functions/Help.Tests.ps1` runs for public functions and is working, but missing a few checks like parameter help
  - `tests/general/Help.Tests.ps1` **would** run for public and internal functions if the `$ModuleName` parameter is updated to `Maester`. Currently it doesn't run anything and would result in many failed tests.
- I'd recommend to update multiple files to [Pester v5 syntax](https://pester.dev/docs/migrations/v4-to-v5). E.g. replace `foreach`-loops with `Describe/Context/It -ForEach @(...)` and place loose discovery code in `BeforeDiscovery`-blocks.
- Avoid dependency on global variables so we can debug tests individually in VSCode without running `pester.ps1` initially.
  - `$global:testroot` is easily replaced with `$PSScriptRoot\..` and rather static
  - `$global:__pester_data.ScriptAnalyzer` could be replaced by tagging the ScriptAnalyzer tests, output failed rules normally in `It 'Should pass ..'` and collect them with `($results.Tests | ? Tag -match 'AnalyzerResult').StandardOutput` in `pester.ps1`

I can offer to update it but would like to discuss which, if any, changes you'd like first.

merill:
@fflaten thanks a lot for the suggestion and apologies for not responding earlier. 

I would definitely take you up on the offer with your suggested improvements.

fflaten:
Partially fixed in #335. Help tests deduplication still remains. Not sure how strict we'd like to be. 
- Only scan public functions?
- Do we need the parameter type validation in `tests/general/Help.Tests.ps1` ?https://github.com/maester365/maester/blob/6ac75f8a4e0c19823e4787d260ca80dad5678f25/powershell/tests/general/Help.Tests.ps1#L110-L136

merill:
@f-bader thoughts? 

I think public functions should be fine.

f-bader:
Second @merill on only public functions

Parameter type validation can be removed in my opinion.

PR #335, opened 2024-07-13
Migrate core tests to Pester v5 syntax

fflaten:
Updates core tests to:
- Use Pester v5 syntax, e.g. no loose code and foreach-loops
- Not depend on global variables. Bad practice and blocks debugging single file in VSCode without running `pester.ps1` once
- Minor improvements
- Rewrites PSScriptAnalyzer.Tests.ps1 to summarize by rule and file. Provides the same visibility in reports, but removes 98% (11k) tests and reduced PSSA-runs to cut time from 20s -> 1s.

Also adds Cmdletbinding-attribute to a few functions missing it. The test validating it was fixed as part of the PR.

Related #98 . Not marking as fixed due to remaining duplicate help-tests


fflaten:
Still ready for review but JUnit export fails in CI for PSScriptAnalyzer.tests.ps1 for some reason (works in devcontainer), so need to troubleshoot it before merge.

f-bader (approved):

PR #340, opened 2024-07-14
Update actions/checkout and action/upload-artifact to node v20 versions

### LINKED ISSUE #339: Upgrade workflow actions to use node v20
Workflows adds a warning about deprecated actions due to use of Node v16. Example:

>[build-validation (macOS-latest)](https://github.com/maester365/maester/actions/runs/9920736780/job/27407929325)
The following actions uses Node.js version which is deprecated and will be forced to run on node20: actions/checkout@v3, actions/upload-artifact@v3. For more info: https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/

fflaten:
Fixes the node v16 and artifact v3 deprecation warnings in workflows.

Hardcodes Maester Test Results report in CI to use Windows-run as `actions/upload-artifacts@v4` doesn't allow overwriting artifacts by default and report should always use the same OS. Windows results are currently used (slow runner so always last upload) and has more PSSA rules, so picked that one.

Fix #339

Only tested build-validation workflows, but checkout-action does not have known breaking changes AFAIK.

### DIFF
--- a/.github/workflows/build-validation-report.yaml
+++ b/.github/workflows/build-validation-report.yaml
@@ -14,7 +14,7 @@ jobs:
     steps:
     - uses: dorny/test-reporter@v1
       with:
-        artifact: test-results
+        artifact: test-results-Windows
         name: Maester Test Results
         path: '*.xml'
         reporter: java-junit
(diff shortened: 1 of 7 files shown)


fflaten:
Report workflow will fail as it executes the workflow in `main`. Verified in fork, see result here: https://github.com/fflaten/maester/runs/27423735756

PR #342, opened 2024-07-14
Updates manifest with correct aliases

fflaten:
Updates aliases in manifest to match current function aliases. It included function names and were missing a few aliases.

Should some of the aliases be removed maybe?


f-bader:
For the time being I would keep the aliases. Thanks for the contributions

PR #351, opened 2024-07-14
Moved additional test error/warning details to verbose

Snozzberries:
This limits the amount of details sent back to the shell with Invoke-Maester without the Verbose switch.


merill (approved):
Agree. This is better and less chatty.

PR #359, opened 2024-07-17
Fixed MS-EXO.2.1

NZLostboy:
This check was not working as expected. It would state that the check passed, and that the safe list is enabled in your tenant. However if the safe list is enabled, the check should fail.

Image showing the issue (prior to changes):
Screenshot 2024-07-17 at 9 51 20 AM

Looking at the other CISA checks, the $null -eq value should actually be -ne.


merill (approved):
Good catch. Thanks for the fix @NZLostboy

PR #371, opened 2024-07-20
Rewrite help and common function tests

### LINKED ISSUE #98: Suggestions to improve module tests

fflaten:
- Deduplicates help tests and expands them to check all exported functions instead of only functions directly in /public/*.ps1.
- Adds missing function description and parameter help to resolve failing tests and fix placeholders in website docs
  - Best effort, copied synopsis for function description. Please review.
- Adds link to online docs for every public function so e.g. `Get-Help Invoke-Maester -Online` will work.
- Fixes test counter in summary for `pester.ps1`. It didn't count tests in `powershell/tests/functions/` resulting in a success message when some tests failed.
- Minor help cleanup/improvements

Related #335 
Fix #98


powershell/tests/functions/Common.Tests.ps1
```diff
@@ -0,0 +1,56 @@
...
+            $functionPath | Should -Exist
+        }
+
+        It "Should be an advanced function" {
+            $function.CmdletBinding | Should -BeTrue -Because 'public functions should be advanced functions'
+            $function.ScriptBlock.Ast.Body.ParamBlock | Should -Not -BeNullOrEmpty -Because 'functions should have a param()-block'
+            $function.ScriptBlock.Ast.Body.ParamBlock | Should -Not -BeNullOrEmpty -Because 'functions should have [CmdletBinding()] attribute for explicit advanced function'
+        }
+
+        # Skipping for Cisa tests until they're updated
```
f-bader, line 32: Would we need this for EIDSCA as well? Or is this not required because we only focus on public functions as of now?

f-bader:
@fflaten Amazing work, thank you so much for your help. Please see latest comment before I merge

powershell/tests/functions/Common.Tests.ps1
fflaten: Public functions for now, so only applicable to `Test-MtEidscaControl` which is fixed.

Should probably create an issue so someone familiar with the Cisa functions can fix them.

Personally I'm not sure the test is needed, but kept it for now. It might force useless verbose output just to be compliant, e.g. `Get-MtSession`.

PR #378, opened 2024-07-22
Additional clarity of permissions for Azure and EXO with App Regs and…

Snozzberries:
… new addition summarizing Container App Jobs


merill (changes requested):
@Snozzberries this is a good docs update. 

In create-entra-app.md should be mention why they are optional (ie call out they are required for running the CISA tests)?

Snozzberries:
> @Snozzberries this is a good docs update.
> 
> In create-entra-app.md should be mention why they are optional (ie call out they are required for running the CISA tests)?

Good thought, updated.

merill (approved):

merill:
Nice. Tx!

PR #388, opened 2024-07-25
Update MT.1006.md

ChrFrohn:
This was a little bit tricky. There was/is difference between the title and description. I have modified the text in the titel, description and How to fix.

The reason for using the phares "user with administrator roles" is because admins are just users with more permissions after all :)

I hope this is all right and it make sense.

Also added Entra admin center link to CA policies and CA policy templates and renamed "learn more" to "Related links"


merill:
Looks good. Thanks a lot @ChrFrohn

PR #389, opened 2024-07-25
Added test MT.1035 for checking RMAU-protection of security groups in Conditional Access Policies

Cloud-Architekt:
- [Added Test] "MT.1035" - All security groups assigned to Conditional Access Policies should be protected by RMAU
- [Enhancement] Added support for receiving result of `Get-GraphObjectMarkdown` as plaintext without format and paragraph


merill (approved):
Awesome work. Looks good!

PR #403, opened 2024-07-31
Add two new inputs - step_summary and artifact_upload

BenStreet365:
Contains additional inputs for disabling step summary and artifact upload, making them optional.


merill (approved):
Thanks for the PR. Yes it's good to have these as options!

PR #405, opened 2024-08-02
New cmdlet to get a local cache of known license SKUs and their nesti…

Snozzberries:
…ng relationships

Still trying to figure out how useful this will be. The current idea is to have an ability to check license relationships. `Get-MtLicenseInformation` uses just the base SKU as the bundles show that entitlement, so this isn't needed, but my thought is it may be helpful to get the relationships as well.

My other concern is I am not certain Microsoft will retain that download URL when updating in the future, would be ideal if it would be an aka.ms link instead.

### DIFF
--- a/powershell/Maester.psd1
+++ b/powershell/Maester.psd1
@@ -123,7 +123,7 @@ FunctionsToExport = 'Add-MtTestResultDetail', 'Clear-MtGraphCache', 'Connect-Mae
                'Test-MtCisaDlp',
                'Test-MtConditionalAccessWhatIf',
                'Test-MtConnection',
-               'Test-MtEidscaControl',
+               'Test-MtEidscaControl', 'Update-MtLicenseCache',
                'Test-MtPimAlertsExists', 'Test-MtPrivPermanentDirectoryRole',
                'Update-MaesterTests', 'Compare-MtTestResult',  'Get-MailAuthenticationRecord',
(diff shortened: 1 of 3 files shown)


merill:
What's the use case/scenario you have in mind for this @Snozzberries?

Snozzberries:
> What's the use case/scenario you have in mind for this @Snozzberries?

It may not be a good fit for Maester, so I can always pull it back to something else. My thinking though was, we are defining the base level a la carte SKU, like Entra ID P2, but there are a few dozen subscriptions that include that. That was the baseline thought, is it possible to define the subscription options, so yeah Entra ID P2 is the requirement but you can attain it through <list>.

This came mainly from digging into the other licenses, like how do I check for a DLP license. 
https://learn.microsoft.com/en-us/office365/servicedescriptions/microsoft-365-service-descriptions/microsoft-365-tenantlevel-services-licensing-guidance/microsoft-365-security-compliance-licensing-guidance#which-licenses-provide-the-rights-for-a-user-to-benefit-from-the-service-7

This is more a nice to have thing too, so no concerns if it is better left out of Maester.

merill:
Got it. Since this isn't a published API, there could be a risk if we build a dependency on this and it changes in the future. I would prefer if we can avoid adding this dependency.

Snozzberries:
No worries, that makes total sense. Closing for now.

PR #410, opened 2024-08-07
Addition of EXO DLP CISA Tests

Snozzberries:
This adds EXO 8.2,3,4 for the CISA DLP tests and a license check.


merill (approved):
Nice @Snozzberries !

Issue #399, opened 2024-07-29
Supress Entra ID Recommendations where required SKU doesnt match the capabilities of the tenant being assessed by Maester


MattWhite-personal:
The Maester Entra ID recommendations test looks at the Active recommendations against a M365 tenant and marks those with a Fail.

![image](https://github.com/user-attachments/assets/867dc804-2136-4999-a9d5-7846f1fd9fc1)

Currently when attempting to follow the remediation action to dismiss the Recommendation the Dismiss / Postpone fails with an error message (I have an open support ticket for this).

To improve the user experience compare the active recommendations against the `requiredLicenses` attribute returned by the test and filter out those that don't match the `subscribedSKUs` against that tenant

merill:
That's a good suggestion. Tx!

PR #414, opened 2024-08-11
Updated CISA Entra tests to use skippedBecause, added tag for minimum…

### LINKED ISSUE #321: Update CISA SCuBA Entra Tests to use skipReason
With the addition of SkipReason, the Entra cmdlets and tests need to be updated to use this rather than the Pester skip.

https://github.com/maester365/maester/blob/main/powershell/public/CISA/exchange/Test-MtCisaAntiSpamAllowList.ps1#L20-L23

https://github.com/maester365/maester/blob/main/tests/CISA/exchange/Test-MtCisaAntiSpamSafeList.Tests.ps1

Snozzberries:
… license level

This is to resolve #321 and the tags will help with #399

### DIFF
--- a/powershell/public/cisa/entra/Test-MtCisaAppAdminConsent.ps1
+++ b/powershell/public/cisa/entra/Test-MtCisaAppAdminConsent.ps1
@@ -18,6 +18,11 @@ function Test-MtCisaAppAdminConsent {
     [OutputType([bool])]
     param()
 
+    if(!(Test-MtConnection Graph)){
+        Add-MtTestResultDetail -SkippedBecause NotConnectedGraph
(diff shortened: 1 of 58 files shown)


merill (approved):
Nice. Yes thanks for adding the tags. I'll approve this but should the tag be without spaces instead of "Entra ID P1"?

Snozzberries:
For the tagging structure. Would it be preferable that they don't have spaces? Unfamiliar with what else may use them, so if it is easier with the filtering in the cmdlets, I am all for it, honestly I had only been thinking about how it would show up in the report and likely didn't consider other dependencies.

image

PR #416, opened 2024-08-13
Proposal to add conditional access gap test

RobbeVandenDaele:
Hi all!

I created a new Maester test to look for conditional access policy gaps. The approach is to look for objects which are excluded in conditional access policies, but not specifically included in at least one other conditional access policy. By doing this, I try to find possibly overlooked exclusions which do not have a fallback. 

I am open for any feedback you have regarding the test 😃 . 

Kind regards
Robbe

### DIFF
--- a/powershell/public/Test-MtCaGaps.ps1
+++ b/powershell/public/Test-MtCaGaps.ps1
@@ -1,0 +1,13 @@
+<#
+.SYNOPSIS
+    Check for gaps in conditional access policies
+
+.DESCRIPTION
+    This function checks if all objects found in policy exclusions are found in policy inclusions.
+
+.EXAMPLE
+    <TO DO>
+
+.LINK
+    <TO DO>
+#>
--- a/powershell/public/Test-MtCaGaps.ps1
+++ b/powershell/public/Test-MtCaGaps.ps1
@@ -11,3 +11,153 @@
 .LINK
     <TO DO>
 #>
+function Get-ObjectDifferences {
+    param (
+        [System.Collections.ArrayList]$excludedObjects,
+        [System.Collections.ArrayList]$includedObjects
+    )
+
+    # Only get unique values
+    $excludedObjects = $excludedObjects | Select-Object -Unique
+    $includedObjects = $includedObjects | Select-Object -Unique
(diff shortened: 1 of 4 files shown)


merill:
Thanks a lot for submitting this as a PR.

I 💛 the idea. I think this will be a good one to have as a  ⚠️ Warning test. Since this is more to highlight accidental issues.Having said that we haven't added the Severity flag to Maester yet.

@f-bader  @Snozzberries  and @Cloud-Architekt  thoughts? Are there suggestions you have to make this better?

Snozzberries:
I do like this, and it has been something I have considered in the past, so glad to see this.

I think splitting it into a few functions and tests would be helpful. Move the helper functions Related and Differences to their own cmdlets, then split user, group, role, app, spn, location, and platform into individual cmdlets or switches and tests.

Having a tag for severity or impact would be valuable too. We could likely add that to the tags as a placeholder until we have a plan to surface in the report.

RobbeVandenDaele:
Hi @merill @Snozzberries 

Thank you for the feedback! I added cmdlets to the two helper functions 'Get-ObjectDifferences' and 'Get-RalatedPolicies', and I added the 'Warning' tag to MT.1036 in 'Test-ConditionalAccessBaseline.Tests.ps1'. 
image
I hope this is what you meant by adding the Warning tag as a placeholder, if not please let me know 😄 .

@Snozzberries I am not sure what you mean by splitting user, groups, roles, etc. into individual cmdlets or tests. Although I do see the value (being able to choose which objects you want to test), I currently fail to see how an end user will be able to choose these flags when using the 'Invoke-Maester' command. 

I am happy to discuss this further. Let me known what is expected from me to help get this test production ready.

merill:
Looks good. I'll merge it in

RobbeVandenDaele:
Thank you for elaborating @Snozzberries! I will try to rework this somewhere in September and create a new PR if done. 

Thank you for merging @merill

RobbeVandenDaele:
@merill @Snozzberries, I see that the documentation at https://maester.dev/docs/tests/maester is not yet updated with the new MT1036 test. I did add the .md file to the website/docs/tests/maester folder. Is there anything else I still need to do?

PR #418, opened 2024-08-15
Updated EIDSCA description and added check AS.04

Cloud-Architekt:
* Updated description and "how to fix" links for EIDSCA JSON
* Added check for SMS phone sign-in (AS4)


merill (approved):
Looks good. Tx!

PR #419, opened 2024-08-16
Fixed Example URLs

NZLostboy:
The folder structure previously changed which broke the good test example URLs. Updated the URLs to point to the correct place.


merill (approved):
Good catch. Thanks a lot @NZLostboy !

PR #420, opened 2024-08-19
Adjust RoleEligibilitySchedule scope

NZLostboy:
I have been doing some work around the CIS checks and came across this. These two tests seem like they might be over scoped, or at least that is what my testing seemed to indicate. Caveat I have an less than ideal test environment at the moment so I want a second opinion.

The original scope was `RoleEligibilitySchedule.ReadWrite.Directory` but in testing `RoleEligibilitySchedule.Read.Directory` seemed to work fine.

Could someone else test and confirm this for me? This would allow us to run these tests without the -Privileged flag

### DIFF
--- a/powershell/public/cisa/entra/Test-MtCisaCloudGlobalAdmin.ps1
+++ b/powershell/public/cisa/entra/Test-MtCisaCloudGlobalAdmin.ps1
@@ -24,9 +24,9 @@ function Test-MtCisaCloudGlobalAdmin {
     }
 
     $scopes = (Get-MgContext).Scopes
-    $permissionMissing = "RoleEligibilitySchedule.ReadWrite.Directory" -notin $scopes
+    $permissionMissing = "RoleEligibilitySchedule.Read.Directory" -notin $scopes
     if($permissionMissing){
-        Add-MtTestResultDetail -SkippedBecause Custom -SkippedCustomReason "Missing Scope RoleEligibilitySchedule.ReadWrite.Directory"
+        Add-MtTestResultDetail -SkippedBecause Custom -SkippedCustomReason "Missing Scope RoleEligibilitySchedule.Read.Directory"
         return $null
     }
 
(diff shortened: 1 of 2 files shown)


Snozzberries:
This issue thread has some good context of where this scope comes into play.

https://github.com/maester365/maester/issues/195

NZLostboy:
> This issue thread has some good context of where this scope comes into play.
> 
> #195

@Snozzberries just to make sure I understand #195 properly. The issue is a user with `RoleEligibilitySchedule.Read.Directory` cannot use `Get-MtRole` to get roles where they're being provided by PIM?

Just wanting to make sure I understand the issue before I test further

Snozzberries:
> > This issue thread has some good context of where this scope comes into play.
> > #195
> 
> @Snozzberries just to make sure I understand #195 properly. The issue is a user with `RoleEligibilitySchedule.Read.Directory` cannot use `Get-MtRole` to get roles where they're being provided by PIM?
> 
> Just wanting to make sure I understand the issue before I test further

Less of an issue with `Get-MtRole` and more with `Get-MtRoleMember`. We are only able to list members of eligible roles with `RoleEligibilitySchedule.ReadWrite.Directory`. We can list active members with just Read though.

NZLostboy:
> > > This issue thread has some good context of where this scope comes into play.
> > > #195
> > 
> > 
> > @Snozzberries just to make sure I understand #195 properly. The issue is a user with `RoleEligibilitySchedule.Read.Directory` cannot use `Get-MtRole` to get roles where they're being provided by PIM?
> > Just wanting to make sure I understand the issue before I test further
> 
> Less of an issue with `Get-MtRole` and more with `Get-MtRoleMember`. We are only able to list members of eligible roles with `RoleEligibilitySchedule.ReadWrite.Directory`. We can list active members with just Read though.

Sorry you are right, `Get-MtRoleMember` is what I meant. Alright I understand the issue, I will see if there is any other way around it otherwise I will probably have to close this this PR.

Seems a little unfortunate to require the write perm for this particular issue.

Thank you for the clarification.

PR #440, opened 2024-08-30
CIS Check First Push

NZLostboy:
First cut for some of the CIS checks.

This is the CIS checks that can be automated (according to CIS, I realise some of them can be done other ways and I will get to that) down to check 1.3.6.

I realise as I am writing this that I have missed the docs/commands entries so I will go back and do those today and add them here.

In the mean time if I could get a review and feedback on whats currently pushed that would be great.


powershell/public/cis/Test-MtCis365PublicGroups.ps1
soulemike, line 1: Limiting this with an upper bound on the number of groups shown in the test result details to like 6 may be necessary to protect from trying to send hundreds or thousands of results.

Example: https://github.com/maester365/maester/blob/main/powershell/public/cisa/exchange/Test-MtCisaSpfDirective.ps1#L89

powershell/public/cis/Test-MtCisCalendarSharing.ps1
soulemike, line 1: It looks like this was a move rather than a copy. So this would break the CISA tests.

powershell/public/cis/Test-MtCisCustomerLockBox.md
soulemike, line 1: Would be worth adding a note that only EXO is currently supported.

powershell/public/cis/Test-MtCis365PublicGroups.md
soulemike, line 1: Under remediation action section bullet 1, setting these to an MD link has been a clean up common for most tests. If there is a  deeper link supported too adding that.

e.g., adding [**Active teams & groups**](https://admin.microsoft.com/#/groups)

powershell/public/cis/Test-MtCisSharedMailboxSignIn.ps1
soulemike, line 1: Would need to add caching through Get-MtExo for L33

https://github.com/maester365/maester/blob/main/powershell/public/cisa/exchange/Get-MtExo.ps1

Snozzberries (commented):
Here are some comments for feedback. Looks like a good start. Thanks!

powershell/public/cis/Test-MtCisCalendarSharing.ps1
NZLostboy: whoops, I will fix that

NZLostboy:
> Here are some comments for feedback. Looks like a good start. Thanks!

Thank you, I will spend some time in the next day or so sorting them all out and adding the other bits.

powershell/public/cis/Test-MtCisCalendarSharing.ps1
NZLostboy: Added it back, will be in the next push

powershell/public/cis/Test-MtCis365PublicGroups.md
NZLostboy: I can definitely make the admin.microsoft.com an MD link, and will add that now. Many of the others are navigation actions and there is no corresponding URL that I could put here. Where there is a deeper link that makes sense, I will add that as well

powershell/public/cis/Test-MtCis365PublicGroups.ps1
NZLostboy: Yeah that makes sense. What about only returning public groups as well? that way we would get more useful results returned?

soulemike: Yep, good catch, returning from `$result` would be better

powershell/public/cis/Test-MtCisSharedMailboxSignIn.ps1
NZLostboy: Added in commit 3ecb4fe5734ee2909c468ebee2bc5dfa60d0cd67

powershell/public/cis/Test-MtCisCustomerLockBox.md
NZLostboy: When you say note, do you mean in the output or a code comment?

soulemike: Just in the MD file, something like this before the Remediation action section

> Test implementation checks Exchange Online service only.

Snozzberries:
This is looking good @NZLostboy, I think the MD file and then passing build tests and this would be good for review. Thank you!

NZLostboy:
> This is looking good @NZLostboy, I think the MD file and then passing build tests and this would be good for review. Thank you!

Thanks @Snozzberries , I have started building out the .mdx files for each test, I assume that is what I am supposed to do with the .mdx files? I am not that familiar with them to be honest.

I will add that note and try to work out the build fails today and then it can go through review.

Snozzberries:
> Thanks @Snozzberries , I have started building out the .mdx files for each test, I assume that is what I am supposed to do with the .mdx files? I am not that familiar with them to be honest.

No need to manually build the mdx those get generated by the pipelines when the PR gets merged.

NZLostboy:
> > Thanks @Snozzberries , I have started building out the .mdx files for each test, I assume that is what I am supposed to do with the .mdx files? I am not that familiar with them to be honest.
> 
> No need to manually build the mdx those get generated by the pipelines when the PR gets merged.

Oh awesome! alright I will work on getting the builds fixed

powershell/public/cis/Test-MtCisCustomerLockBox.md
NZLostboy: Added in commit 54cf74edeef7223295fdfaa3d453f28160a72f49

NZLostboy:
Builds fixed, sorry for the number of builds. I didn't know how to perform the tests locally but now I do.

merill (approved):
Nice. This is looking good!

Issue #447, opened 2024-09-03
Get-MTUser Does Not Query Guest Users


lbrauns:
The function ``Get-MTUser`` allows the type ``Guest`` as UserType. This value is never processed and the function is currently not able to query guest users. Instead a ``Get-mtuser -UserType Guest`` will return member users.

I will open a pull request that fixes this.

merill:
Fixed. Please retest with -preview module (or on main once published).

PR #448, opened 2024-09-03
Implements Query for UserType "Guest" (maester365/maester#447)

lbrauns:
Changes the filter for the Graph query correctly if UserType "Guest" was supplied.
Error handling for the impossible case that another UserType than "Member" or "Guest" is supplied is added as well.


merill (approved):
Cheers. Thanks for the fix!

Issue #453, opened 2024-09-05
MS.AAD.8.2 will fail if allowInvitesFrom is set to none


mattcave:
MS.AAD.8.2 explicitly checks the authorizationPolicy for the allowInvitesFrom value being set to "adminsAndGuestInviters" (and nothing else). If the value is set to "none", the test will fail. Do you agree that "none" should also be seen as compliant?

Snozzberries:
@mattcave thanks for this suggestion and great note. I would agree. If you feel comfortable doing the PR, looks like adding or none to the test would be a great PR.

https://github.com/maester365/maester/blob/main/powershell/public/cisa/entra/Test-MtCisaGuestInvitation.ps1#L28

mattcave:
Proposed fix here - https://github.com/maester365/maester/pull/454

merill:
Cheers. @mattcave merged!

PR #454, opened 2024-09-08
Update MS.AAD.8.2 test to allow "none"

mattcave:
Fix for https://github.com/maester365/maester/issues/453, "MS.AAD.8.2 will fail if allowInvitesFrom is set to none"


merill (approved):
Great PR and thanks for the fix.

PR #464, opened 2024-09-10
Migrate MtCisaContactSharing to use Get-MtExo caching

NZLostboy:
MtCisaContactSharing was calling Get-SharingPolicy instead of using the Get-MtExo cached version. Migrated to using the cached version for consistency and speed.


merill (approved):
Nice!

PR #475, opened 2024-09-20
CIS Checks 2.1.1, 2.1.2, 2.1.3, 2.1.4

NZLostboy:
CIS checks down to check 2.1.4


NZLostboy:
Realized these need to consider the default policy as that’s what the CIS check basically looks for. Not sure how these checks will work with multiple polices so I’ll test and update

merill:
These are looking good @NZLostboy 

Let me know when you want me to merge them in. Cheers!

merill (approved):
Looks good. Awesome work @NZLostboy!

PR #480, opened 2024-09-25
Update azure-automation.md

BurgerhoutJ:
Added line 108 Update-MaesterTests to get the latest tests

### DIFF
--- a/website/docs/monitoring/azure-automation.md
+++ b/website/docs/monitoring/azure-automation.md
@@ -105,6 +105,7 @@ cd $env:TEMP
 md maester-tests
 cd maester-tests
 Install-MaesterTests .\tests
+Update-MaesterTests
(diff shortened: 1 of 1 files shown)


merill:
@BurgerhoutJ have you run into an issue with stale tests? The Install-MaesterTest calls the same function as Update-Maester the only difference is the check for existing folder.

So adding this extra command is not going to make any new changes.

BurgerhoutJ:
Strange. When I added the update-maestertests, after the run I had 240+ tests instead of the 135 the first few times. 

I will check it again with and without the added command.

merill:
Were you able to repro the issue @BurgerhoutJ ?

BurgerhoutJ:
Sorry. Didn't have the time this week to repro.

You can close this pull request.

Issue #463, opened 2024-09-10
Test-MtEidscaAM02 Checks For Wrong Value


lbrauns:
Test ``Test-MtEidscaAM02`` us supposed to check if Authenticator OTP is enabled but checks for the value ``.state``. This is the same as ``Test-MtEidscaAM01``. Instead the test should check for ``.isSoftwareOathEnabled`` -eq 'true'

Issue #459, opened 2024-09-09
Test-MtEidscaPR05 fails because $testResult is set to type System.String


m-hanisch:
### Reproduce error
Test [Test-MtEidscaPR05.ps1](https://github.com/maester365/maester/blob/main/powershell/internal/eidsca/Test-MtEidscaPR05.ps1) fails because $tenantValue is set to type System:String `[string]$tenantValue` which results in the a wrong test result of the `-ge` [Comparison Operator](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_comparison_operators?view=powershell-7.4#-gt--ge--lt-and--le).

The description states:
> The minimum length in seconds of each lockout. If an account locks repeatedly, this duration increases.
> 
> Queries settings
> and returns the result of
> graph/settings.values | where-object name -eq 'LockoutDurationInSeconds' | select-object -expand value -ge '60'

[lines 25 to 28](https://github.com/maester365/maester/blob/061f1487cb65d994ef1ca3e6ead5ee0877c289f1/powershell/internal/eidsca/Test-MtEidscaPR05.ps1#L25-L28)
```powershell
PS C:\Users> $result = Invoke-MtGraphRequest -RelativeUri "settings" -ApiVersion beta
PS C:\Users> [string]$tenantValue = $result.values | where-object name -eq 'LockoutDurationInSeconds' | select-object -expand value
PS C:\Users> $tenantValue
120
PS C:\Users> $testResult = $tenantValue -ge '60'
PS C:\Users> $testResult
False
# 120 is greater than 60.
```

Moreover, the value not null test seems to be static/hardcoded.

[line 29](https://github.com/maester365/maester/blob/061f1487cb65d994ef1ca3e6ead5ee0877c289f1/powershell/internal/eidsca/Test-MtEidscaPR05.ps1#L29)
```powershell
$tenantValueNotSet = $null -eq $tenantValue -and '60' -notlike '*$null*'
```

### Issue resolution
```powershell
PS C:\Users> $testResult = [int]$tenantValue -ge [int]'60'
True
```

```powershell
$tenantValueNotSet = $null -eq $tenantValue -and $tenantValue -notlike '*$null*'
```

Issue #451, opened 2024-09-05
Test-MtEidscaCR02 And Test-MtEidscaCR03 Check For Same Setting


lbrauns:
The test ``Test-MtEidscaCR02`` (Checks if Consent Framework - Admin Consent Request - Reviewers will receive email notifications for requests is set to 'true')  and ``Test-MtEidscaCR03`` (Checks if Consent Framework - Admin Consent Request - Reviewers will receive email notifications when admin consent requests are about to expire is set to 'true') both check the value ``notifyReviewers``.

The test ``Test-MtEidscaCR03`` should check instead for the value of ``remindersEnabled``.

PR #487, opened 2024-10-06
Bug fixes on invalid or wrong EIDSCA value checks

Cloud-Architekt:
* Fix: [Test-MtEidscaAM02 Checks For Wrong Value](https://github.com/maester365/maester/issues/463)
* Fix: [Test-MtEidscaCR02 And Test-MtEidscaCR03 Check For Same Setting](https://github.com/maester365/maester/issues/451)
* Fix: Temporary fix on [Test-MtEidscaPR05 fails because $testResult is set to type System.String](https://github.com/maester365/maester/issues/459) will be overwritten by next build process on EIDSCA checks. Therefore a logic to identify integer value for comparison has been added to the `Update-EidscaTests.ps1` script.


merill (approved):
Thanks @Cloud-Architekt !

Issue #427, opened 2024-08-26
No artifact recieved after update 0.3.0


Haakonak:
Hello Maester community! I have a question/problem with artifacts. After updating to the latest version 0.3.0 i no longer recieve an artifact after running Maester as an action.

I am using the given Maester as an Action script, with the modification of adding "mail_userid" and "mail_recipient".
```
jobs:
  run-maester-tests:
    name: Run Maester Tests
    runs-on: ubuntu-latest
    steps:
    - name: Run Maester action
      uses: maester365/maester@main
      with:
        client_id: ${{ secrets.AZURE_CLIENT_ID }}
        tenant_id: ${{ secrets.AZURE_TENANT_ID }}
        include_public_tests: false # Optional 
        pester_verbosity: 'None' # Optional - 'None', 'Normal', 'Detailed', 'Diagnostic'
        mail_userid: 'example.userid'
        mail_recipients: 'example.user@mail.com'
```

From the run logs, the parameter "artifact_upload" returns "true", but when looking through the summary there is no artifact. 
I have attempted adding the "artifact_upload" parameter and manually setting it to "true" without any results. I have also tried to remove "mail_userid" and "mail_recipients" but still no artifact. 

If anyone has any tips or info that could explain this issue to me i would be very grateful. If this does not qualify as an issue, i can post it in the Discussions forum instead.

merill:
Do you see the .html file being created?

Haakonak:
I compared my latest Maester run with a previous one which did produce an artifact, and the "actions/upload-artifact@v4" is skipped completely. However, right before displaying how many tests were run, passed and failed, it does seem like the .html is being created.

```
🔥 Maester test report generated at test-results/test-results.html

Tests Passed ✅: 74, Failed ❌: 39, Skipped ⚫: 22

Script execution Complete
```

merill:
@f-bader Could this be an issue with Github Action not uploading the artifact?

f-bader:
Will check if there is a fail or something that prevents this. Should not happen

f-bader:
Could this be related to https://github.com/actions/upload-artifact/releases/tag/v4.4.0

> We will no longer include hidden files and folders by default in the upload-artifact action of this version. This reduces the risk that credentials are accidentally uploaded into artifacts. Customers who need to continue to upload these files can use a new option, include-hidden-files, to continue to do so.

Haakonak:
This does seem like it could be whats causing this issue. 

[...]

PR #493, opened 2024-10-10
Added parameter to upload html artifact

Haakonak:
Refering to issue #427 

I have never contributed to open source before, so if this is the wrong way to do it, please let me know :)

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -146,3 +146,4 @@ runs:
       with:
         name: maester-test-results-${{ env.NOW }}
         path: test-results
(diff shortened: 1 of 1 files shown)


merill (approved):
I think this should do it. Thanks a lot @Haakonak and congratulations on your first PR contribution. 👏

FrederikLeed:
Did you validate the fix solving the issue? It seems like actions/upload-artifact@v4 is never actually executed and thus the fix not having any value?

PR #498, opened 2024-10-15
Fixed CIS Policy Checks

NZLostboy:
Reconfigured the CIS based policy checks to specifically check the default policy. Prior to this I am not 100% what the outcome of the existing logic would have been which is an oversight. It likely only considered the first policy listed.

Checking the default policy is typically how CIS provides automated guidance. Where only the default policy is checked, it states this in the title of the output for clarity.

In the future we can probably make this better and check all policies for the desired outcome, but for now I think the clarity of the check is more important.


merill (approved):
Tx, yes it's good to call it out that we check just the default policy.

Issue #484, opened 2024-10-03
Delegation for Connect-Maester


slueders-bag:
it would be great as an CSP to connect to different tenants by using an parameter like TenantName oder TenantID in the Connect-Maester cmdlet.

Snozzberries:
This would be nice, but becomes challenging since each of the upstream modules uses a different parameter structure and approach. In my experience it is easier to connect through those upstream commands then skip Maester connection in multi-tenant scenarios. [example](https://maester.dev/docs/monitoring/azure-container-app-job#:~:text=Connect%20to%20target%20tenant)

Considering other modules like PNP.PowerShell and handling that as a custom MSAL wrapper, I feel that has other challenges that make it even less desirable for multi-tenancy.

M365DSC team has a nice table that shows some of the complexity. https://microsoft365dsc.com/user-guide/get-started/authentication-and-permissions/#authentication-methods

merill:
We added Connect-Maester as a convenience for the 80% use case and ideally don't want to duplicate all the functionality in Connect-MgGraph in Graph and similar commands in the other modules.

However I think it will be good for the Maester docs to include guidance on how to use Connect-MgGraph and other commands to switch tenants etc, (and point to the docs for the various modules).

PR #505, opened 2024-10-22
New Connecting Details Docs Page

Snozzberries:
Questions around `Connect-Maester` and authentication to support tests in general has come up a few times. Added a new docs page to attempt to address some additional guidance.

@merill, Unsure if the React deployment can support the Mermaid diagram.
https://docusaurus.io/docs/next/markdown-features/diagrams

Ref #484


merill (approved):
This is a good start. Thanks @Snozzberries 

I added a basic section (which is mainly what we have in the command reference doc) and created an advanced page with the details you had.

PR #509, opened 2024-10-25
Beta of ORCA

soulemike:
Known issues:

- [x] Get-MtExo requires additional command support. Shown below.
- [x] Get-ORCACollection is modified to use Get-MtExo to introduce caching, but it is still called in each test rather than just the necessary policy lookup.
- [ ] There are a couple EXO cmdlets that are not set to cache in Get-ORCACollection due to parameters.
- [x] Add-IsPresetValue uses Add-Member and due to calling Get-ORCACollection, this causes errors to be shown

Priority known issues:

- [x] Add-MtTestResultDetail would benefit form adding in the base MD files.
- [x] The test is executed from the Pester test file and this probably isn't the most desirable.


Snozzberries:
@merill Plenty of improvements to still do on this, but wanted to validate this approach will work in prerel if you are open to that.

cammurray:
Hey! Checking in here, is this functional? @merill  @soulemike

soulemike:
> Hey! Checking in here, is this functional? @merill @soulemike

Hey Cam, was just chatting with Merill on this. Was waiting for some of the v1 announcement stuff to come through, but will likely start getting this ready soon. I want to test this in the prerel and then get working on merging the docs and remediation stuff in before the end of year.

soulemike:
@merill once you are back from your travels this should be good to go.

merill:
Looks good @soulemike 

Thanks a lot, this is amazing work.

We do need to review the tests and add markdown as well as fix the ones where the details shows failure but the test passes.

fflaten:
Great work on this! I'm a little late here, but quick question. Why does every internal check file include the definitions for base class and enum? Aren't they all the same, meaning we're publishing 19k lines and also executing them 58x extra during module import.

soulemike:
@fflaten, you are correct. I couldn't get the definitions to work in a standalone file with using or similar, or without making the module more dependent on these optional features. So I just back burned it with the workaround. It can certainly be improved, but I didn't see an easy way in the moment. 🙃

PR #515, opened 2024-10-29
Update Test-MtConditionalAccessWhatIf.ps1

milanschwartz:
-Added DeviceInfo, InsiderRiskLevel,ServicePrincipalRiskLevel params
-Added IncludeReportOnly,IncludeDisabled,PrettyJsonVerboseOutput switches
-Using Test-MgBetaIdentityConditionalAccess cmdlet
-Some Verbose logging


merill (approved):
@milanschwartz Thanks a lot for your patience!

This is an excellent PR with lots of code samples and nicely done.

I did make one minor edit. In Maester we avoid dependencies of Graph PS cmdlets and instead prefer invoking the REST API directly with Invoke-Mg / Invoke-MtGraphRequest.

This way we don't need to deal with dependencies and changes to the PowerShell module.

PR #518, opened 2024-11-04
Add Connection to Teams and the first Meeting Policies Tests

weyCC81:
**Add Connection**
I have added the necessary test to connect to Teams without yet impacting the "All" scope.
In relation to `-SkipGraphConnect` I have adjusted some dependencies which need validation (see ToValidate and ToAdjust)

**Teams Meeting Policies**
I have implemented the first Test, which are also in the M365 Secure Score for the area Teams.

PS: As this is one of my first commits, I am really open to feedback


merill:
Thanks for this @weyCC81 

One of the assumptions we make is that the user will not have the other modules like Teams and Exchange installed. So the part where Test-Cs... is called needs to be wrapped in a try..catch so it doesn't throw errors for users who don't have the module installed.

weyCC81:
Does the Build Error have to do something with me or:
- https://github.com/maester365/maester/commit/5d18a04d2604f2d85cdc993610fc6476849bcf39
- https://github.com/maester365/maester/pull/539

merill:
No you can ignore that build error.

merill:
Thanks for the update @weyCC81 

I've made some more updates after doing a few tests.

I think it's starting to get there. I'm going to go ahead and merge into main, but will open an issue to keep track of and address these to make it easier for users to follow the test and the recommendation.

1. Move each test into Maester cmdlets
2. Provider a reference number for each test e.g. MT1xxx (like we do with the other Maester tests)
3. Include the reference number as a tag
4. Improve the markdown to give guidance on the steps to get to the blade to make the change (including the descriptive label) and the Recommended Action step.

PR #530, opened 2024-11-09
Add Exchange Settings Tests

weyCC81:
**Exchange Settings**
I have extended the Tests which are not related to EIDSCA or CISA, which are also in the M365 Secure Score for the area exchange.


merill:
@weyCC81 Thank you so much for this really good PR, and I'm super sorry on the delay in getting back to you on this.

Would it be possible to move the code for this test into a cmdlet instead of the tests file. You can also move the markdown into .md files (see the code samples for one of the other tests).

This ways its easier to make updates/bug fixes to the module and push them through instead of overwriting any changes users might make to their file.

merill:
I've added numbering to the tests. Will include these along with the other set of tests where we want to move them to PowerShell modules and improve the docs/guidance

PR #540, opened 2024-11-14
Update Test-MtCaMfaForAdminManagement.ps1

DanielWahlgren:
Add check for includeApplications - All


merill (approved):
Good catch. Thanks @DanielWahlgren !

PR #560, opened 2024-12-06
fix typos in Run Maester action variables

wesmacdonald:
Update of GitHub Workflow Sample


merill (approved):
Good catch. Thanks @wesmacdonald

PR #566, opened 2024-12-09
Fix failed action

nathanmcnulty:
One too many parenthesis

image


merill (approved):
Ouch. Thanks @nathanmcnulty!!

Issue #569, opened 2024-12-10
Maester - Teams Tests - Fixes


merill:
This issue is to track the updates planned for the Maester - Teams tests cc @weyCC81 

1. Move each test into Maester cmdlets
2. Provider a reference number for each test e.g. MT1xxx (like we do with the other Maester tests)
3. Include the reference number as a tag
4. Improve the markdown to give guidance on the steps to get to the blade to make the change (including the descriptive label) and the Recommended Action step.

merill:
@weyCC81 I've added references and updated the first Teams test as a template.

* Added Test-MtTeamsRestrictParticipantGiveRequestControl
* Added .md file
* Instructions on how to fix

Would you be able to update the rest of the tests to follow the format.

PS: A future enhancement would be to add the Teams related CIS tests. Maybe @NZLostboy has thoughts. I think some of these are very similar.

NZLostboy:
@merill Yeah absolutely, sorry crazy end to the year and I haven't had time to finish the CIS checks.

I have some time put aside after next week to power through and finish all the CIS checks. Where we can combine the Teams tests into one test that makes total sense, but I am trying to keep the CIS tests representative of the documentation where possible so its a balance. Caching them where we can would be the bare minimum, so we can definitely achieve that. 

I can pick up the state end next week when things settle down and get stuck in. Happy to collaborate here or Discord to work together on this

merill:
Agree, we do need to keep the rules isolated so the CIS ones can track against the CIS definitions. 

I would want to avoid adding the same test to the Maester core tests if they are a simple dup.

Issue #521, opened 2024-11-06
Multiple tests fail to run correctly when tenant has a privileged Group with access into the tenant via a GDAP relationship


mattcave:
We work with a number of tenants who have GDAP relationships with multiple vendors (e.g. a vendor that they buy MS licences from). If that GDAP relationship includes the role **Global Administrator**, a number of tests will fail to run correctly. This is because the call to `Get-MtRoleMember` will result in a 404 error - the code tries to make a Graph call to https://graph.microsoft.com/v1.0/groups/<group_id>/members but that call (correctly) returns a 404 as that object is not part of the target tenant. 

I'm not sure what the correct behaviour is in this scenario. 

* Having a Global Admin role on the GDAP relationship is not a good practice, but it's allowed and shouldn't cause the test to throw an error.
* That said, if we cannot pull the details for that group, the test cannot pass because it won't have the information to make a determination.

Perhaps the best course of action is to provide better info about why the test is failing (ie we can't pull all the info we need)?

I believe this affects all MS.AAD.7.x tests

soulemike:
MS.AAD.7.1 with Direct and Indirect reseller relationship setup, without any role delegations (i.e., Legacy DAP). The tests run successfully.
![image](https://github.com/user-attachments/assets/14be4d9f-b90a-46cc-bef3-5733373f8ecd)
![image](https://github.com/user-attachments/assets/067f4b32-abf4-4385-a0e6-f993e0f172ba)

Will get access to generate a GDAP relationship for my test environment later and test again, as that will likely be the culprit. Just documenting different scenarios to identify scope.

mattcave:
![Screenshot 2024-11-06 at 11 08 40 AM](https://github.com/user-attachments/assets/7c89c756-5def-4b19-805c-9327572e9e8d)
Example of a configuration that we see causing an error, to help you with reproducing the issue.

soulemike:
A test with Global Reader

![image](https://github.com/user-attachments/assets/aae8eb54-1d76-42fb-89cd-4cdecae8614a)
![image](https://github.com/user-attachments/assets/ea954e46-cac0-4037-b10a-56a16769b5cf)


A test with Global Admin

![image](https://github.com/user-attachments/assets/b565951e-45e3-475a-9cc6-ee6f9a377dd7)
![image](https://github.com/user-attachments/assets/21b4425b-5a92-4d87-b6b7-3433b309cda5)


@mattcave do you have a few more examples on maybe narrowing down this?

[...]

PR #572, opened 2024-12-10
Cleanup in Prep for v1

soulemike:
- [x] Teams not connected link goes through to Install Guide and not connect-Maester
- [x] Add list of modules to Connect-Maester docs
- [x] CIS 2.x aren't prepended with CIS
- [x] Prepend Teams tests with MTxxx - #569 Added a placeholder
- [x] AAD.7.1,2,3,EXO.16.1,17.3,8.1,2,4 no details on failure
- [x] EXO.1-9 - Prepend a 0 for sort order


soulemike:
The AAD errors were produced due to the scenario described in #521. With this fix that issue should be addressed as well.

PR #577, opened 2024-12-12
Rename happend for this menu - Group owner consent settings have been…

michaelmsonne:
Typo - Removed and replaced with Team owner consent settings. See more here: https://aka.ms/TeamRscSettings


merill (approved):
Thanks for the fix @michaelmsonne!

PR #579, opened 2024-12-12
Add missing Entra ID roles added sense last change

michaelmsonne:
See more here about the rule: https://learn.microsoft.com/en-us/entra/identity/role-based-access-control/permissions-reference


merill (approved):
Thanks!

PR #585, opened 2024-12-13
Spelling and whitespace fixes

SamErde:
A little bit of polish for Maester:

- Add cSpell settings with project word list and ignored words
- Import cSpell dictionaries for PowerShell and C# projects
- Fix multiple typos
- Remove trailing whitespace and lines that only contain extraneous whitespace

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -1,0 +1,33 @@
+{
+  "import": [
+    "@cspell/dict-powershell/cspell-ext.json",
+    "@cspell/dict-csharp/cspell-ext.json"
+  ],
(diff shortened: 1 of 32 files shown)


merill (approved):
Awesome. Thanks @SamErde 

for future reference the EIDCSA.ps1 files are auto generated so any changes will be overwritten when we run the script in ./build

SamErde:
Ah, thanks for the reminder! I would like to know more about how that part of the build works. I have also been curious about why they're in a different folder structure than the other tests.

Issue #549, opened 2024-11-25
Add support for .co.uk and uk.com domains


chriswhitingBCH:
These are both tLDs and should not be tested for SPF, DMARC, etc. 

![image](https://github.com/user-attachments/assets/4845d57b-43c0-40d6-8fad-df70d0950f0c)

chriswhitingBCH:
![image](https://github.com/user-attachments/assets/6ded51c5-f2b8-4a6f-844b-342457d1dd05)

chriswhitingBCH:
Exclude the onmicrosoft.com too 

![image](https://github.com/user-attachments/assets/1248d389-ca59-4f05-b9ea-b609b1d2f24c)

soulemike:
This is an interesting topic for a couple of reasons:
1) What is a technical implementation that allows for filtering well-known managed second-level domains? I am not aware of an organization **restricting** this, just multiple that maintain well-known domains within their own standards.
2) Technically this is still accepted risk. Microsoft **should** have parked records for onmicrosoft.com, the UK **should** have parked records for those top-level and second-level domains. Otherwise your organization has some implicit, al be it very little, trust and thus risk that those could be spoofed.

Open to suggestions on how to identify the first elegantly and then potentially flag as a warning for the second.

merill:
Since the admin can't do much here to fix TLDs it might be best to avoid showing this as failed. 

We could still list them in the table here and flag as ignored.

I think we can hard-code the domains in the PowerShell cmdlet and folks can submit PRs to add them.

We should also add support to pass in domains to be ignore for this test. When we add support for configuring parameters (hopefully this quarter), admins will be able to exclude other domains.

Thoughts @soulemike?

soulemike:
> Since the admin can't do much here to fix TLDs it might be best to avoid showing this as failed.
> We could still list them in the table here and flag as ignored.

This sounds like a reasonable approach. Only use the domains explicitly registered with Entra for the pass/fail, then show the results for the hierarchy but as ignored for the overall result.

> We should also add support to pass in domains to be ignore for this test. When we add support for configuring parameters (hopefully this quarter), admins will be able to exclude other domains.

Yep, allowing an ignored domains array as a property makes sense as well.

I can work on wireframing these two changes up.

PR #599, opened 2024-12-23
Fix a couple bugs in record lookups and allowed for a CISA Strict par…

soulemike:
…ameter for DMARC

Ref #549


merill (approved):
Thanks Mike!

PR #601, opened 2024-12-23
Update error message when DNS record cannot be found

deleugpn:
I recently worked on installing `DNSClient-PS` on the environment which I use to run `maester` so that some of the ExchangeOnline tests can properly work. One tricky thing that I faced while working on this is that the general error message and the error message when a record isn't located is exactly the same, however not finding a record is actually a "successful" execution with a failed result while the general exception is a complete failure to execute. Making these error message different would help diagnose the state of the system and the result of the tests.


merill (approved):
Cheers. This makes sense. Thanks for the PR @deleugpn!

Issue #594, opened 2024-12-18
`MS.AAD.6.1` passes when at least one custom domain is configured correctly


jed-exotic:
https://github.com/maester365/maester/blob/d67de01cd7286e4207a9fa6fdcef5b646517247c/powershell/public/cisa/entra/Test-MtCisaPasswordExpiration.ps1#L44

The scenario is that there are multiple custom domains in an Entra ID tenant. One is unverified (for edge case testing), and two are verified. 

The code implies that at least 1 custom domain must be configured correctly. 

What is the rationale in only evaluating compliance for one of the results that match the test? Is this intentional, or an oversight?

jed-exotic:
I see in the comments above (on the latest commit):
https://github.com/maester365/maester/blob/main/powershell/public/cisa/entra/Test-MtCisaPasswordExpiration.ps1#L11
> Returns true if at least 1 domain has password expiration of 100 years or greater

So this appears to be the expected behavior/implementation, however I would like to understand the justification for not verifying all domains.

jed-exotic:
This can be adjusted to be a stronger test by modifying the `$testResult` accordingly:

> $testResult = ($result | Measure-Object).Count - ($managedDomains|Measure-Object).Count -eq 0

jed-exotic:
Alternatively, if the intention is to only check the Primary/Default domain we may disregard the previous suggestion and instead supply an additional condition here:

```
    $managedDomains = $result | Where-Object {`
        $_.authenticationType -eq "Managed" -and `
        $_.PasswordValidityPeriodInDays -ge 36500 -and `
        $_.isDefault
    }
```

merill:
@jed-exotic thanks for the discussion and the PR, I believe the implementation is based on SCUBA Gear but I agree that we should check all managed domains are not doing password expiry to have this rule pass.

In the PR you are filtering out managed domains that are not verified, I believe they should be included in the check too (e.g. there could be a tenant with no verified domains).

Are you okay to update the PR?

@soulemike Let us know if you think otherwise.

jed-exotic:
> In the PR you are filtering out managed domains that are not verified, I believe they should be included in the check too (e.g. there could be a tenant with no verified domains).


I'm not super familiar on the topic of domain verification in Entra, but I don't believe you can configure password expiry on an unverified domain. If that's a possible configuration, then I can incorporate the suggestion.

merill:
Moving to the backlog as enhancement

PR #603, opened 2024-12-26
fix(MS.AAD.6.1): password expiration must be configured for all domains

jed-exotic:
addresses https://github.com/maester365/maester/issues/594

This may be an opinionated change, and I am happy to discuss.

To comply with MS.AAD.6.1, all verified domains should be evaluated for this configuration value. 

If configuring the primary domain is sufficient for compliance (i.e. the configured value for non-primary domains becomes irrelevant), then the `isDefault` attribute should be used to filter the results down to a single domain.

This PR is with the following requirement in mind:

**ALL** verified managed domains shall be configured to not require password expiry.

### DIFF
--- a/powershell/public/cisa/entra/Test-MtCisaPasswordExpiration.ps1
+++ b/powershell/public/cisa/entra/Test-MtCisaPasswordExpiration.ps1
@@ -38,11 +38,13 @@ function Test-MtCisaPasswordExpiration {
     #$federatedDomains = $result | Where-Object {`
     #    $_.authenticationType -ne "Managed"}
 
-    $managedDomains = $result | Where-Object authenticationType -eq "Managed"
+    $verifiedDomains = $result | Where-Object isVerified
+
+    $managedDomains = $verifiedDomains | Where-Object authenticationType -eq "Managed"
 
     $compliantDomains = $managedDomains | Where-Object PasswordValidityPeriodInDays -ge 36500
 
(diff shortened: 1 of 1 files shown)


weyCC81:
In my opinion, it's a good practice to check all domains, as the 'primary domain' does not always technically appear as 'isDefault.'

merill:
Added comment https://github.com/maester365/maester/issues/594#issuecomment-2574317728

merill:
Holding PR to update test to align with suggestion from @soulemike (see link above).

I agree more details here would be good. Likely returning each domain in a table with the configuration and pass state, similar to other tests. Something like the below. With the test result validating all verified managed and federated domains are valid.

| Domain (Default) | Verified | Type | Validation |
| - | - | - | - |
| contoso.com (✔️) | Verified | Managed | ✅ Pass |
| sub.contoso.com () | Verified | Managed | ❌ Fail |
| contoso.onmicrosoft.com () | Verified | Managed | ✅ Pass |
| oldbrand.com () | Verified | Federated | 🗄️ Skipped |
| google.com () | Unverified | Managed | 🗄️ Skipped |

Reviewing the CISA REGO it looks like they do something similar and measure the aggregate, so just looks like something I missed when wiring this up. https://github.com/cisagov/ScubaGear/blob/main/PowerShell/ScubaGear/Rego/AADConfig.rego#L681

soulemike:
Updated PR to include details.

merill (approved):
LGTM

PR #605, opened 2024-12-31
Update Test-MtConditionalAccessWhatIf.mdx

lukasgd1913:
In the example, an @ is missing before the {} of the 'DeviceInfo' parameter. This causes a syntax error. The example has been corrected in this PR.


merill (approved):
Good catch. Thanks for the PR @lukasgd1913

PR #615, opened 2025-01-08
Respect value of install_prerelease for Maester module in the action

wmacdonald10:
The if condition always evaluated to false when install_prerelease was true or false as it was only checked for being null. Write out to the log file whether the latest or prerelease module was used.


merill (approved):
Cheers thanks for the fix @wmacdonald10!

Sorry for the delay in merging.

Issue #606, opened 2025-01-02
GitHub Action Error: The 'Invoke-MtGraphRequest' command was found in the module 'Maester', but the module could not be loaded due to the following error: [Assembly with same name is already loaded]


robm82:
I have set up a GitHub action to automate the generation of the Maester reports, which do get generated, but I see several errors in the console log:

![image](https://github.com/user-attachments/assets/708e4d6d-c1bf-4ef4-89e5-da9ff3ef9c70)

I did some research into the error, and I'm not sure what the fix is, but someone mentioned that when installing the module as part of the action.yml, to use -AllowClobber.

Any help is greatly appreciated, and awesome work with Maester 🤓

wmacdonald10:
I have the same issue, been trying to track down the cause with no success yet :(

wmacdonald10:
I was finally able to reproduce the error in Ubuntu on WSL.

1. I installed PowerShell Core
2. Installed and ran Maester (Invoke-Maester) 
3. Did not get the Invoke-MtGraphRequest "Assembly with same name is always loaded" errpr
![image](https://github.com/user-attachments/assets/bc677cbe-7bd3-4c34-915f-582cfe2e72cc)



4. Installed Microsoft.Graph (installs all latest modules v2.25.0
5. Run Invoke-Maester

![image](https://github.com/user-attachments/assets/18139e94-fb16-4c21-bc0f-c3de9a788a46)

wmacdonald10:
I tested the following workaround #619 to get around the error

wmacdonald10:
[celebrate]     Wes MacDonald reacted to your message:
________________________________

PR #619, opened 2025-01-13
Workaround for Invoke-MtGraphRequest assembly error on GitHub-hosted runner

wmacdonald10:
The 'Invoke-MtGraphRequest' command was found in the module 'Maester', but the module could not be loaded due to the following error: [Assembly with same name is already loaded]

Brute force removal of offending module version (required by Maester)

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -103,6 +103,10 @@ runs:
             Install-Module Maester -Force
           }
 
+          # Delete Microsoft.Graph.Authentication 2.9.1 (issue #606)
+          Write-Host "Delete module version 2.9.1 of Microsoft.Graph.Authentication"  
+          rm -r /home/runner/.local/share/powershell/Modules/Microsoft.Graph.Authentication/2.9.1
+
           # Configure test results
           $PesterConfiguration = New-PesterConfiguration
           $PesterConfiguration.Output.Verbosity = '${{ inputs.pester_verbosity }}'


action.yml
```diff
@@ -103,6 +103,10 @@ runs:
             Install-Module Maester -Force
           }
 
+          # Delete Microsoft.Graph.Authentication 2.9.1 (issue #606)
+          Write-Host "Delete module version 2.9.1 of Microsoft.Graph.Authentication"  
+          rm -r /home/runner/.local/share/powershell/Modules/Microsoft.Graph.Authentication/2.9.1
```
SamErde, line 108: It may help avoid future errors and workflow failures if you wrap the `rm` command in an `if` statement that checks if the *2.9.1* directory is already there before trying to remove it.

wmacdonald10: I did not expect this to become a permanent solution. It would be much better if the module's dependencies on Microsoft.Graph.Authentication were updated to avoid this assembly collision.

merill:
@wmacdonald10 is this still an issue without your fix? I couldn't repro and haven't seen any issue reported related to this.

wmacdonald10:
> @wmacdonald10 is this still an issue without your fix? I couldn't repro and haven't seen any issue reported related to this.

I just ran the action again from my repo referencing your copy of the action and while the result is marked as green (success), it still throws the error reported in issue #606  during the run. If you want the logs of the action run from my pipeline let me know.

image

jamie-oconnell:
Am also seeing this issue with a fresh repo using the same example workflow given in the README.

merill (approved):
Approving temporary fix.

merill:
I've no idea why but GitHub is not letting me merge this in. Might need to create a new PR for this.

merill (approved):

merill:
@jamie-oconnell I've pushed this through as #673 

Can you please verify if that fixed it? Will close this PR.

jamie-oconnell:
@merill yes thanks! This looks to have worked around the issue. Not seeing the errors any longer on new runs.

merill:
Tx to @wmacdonald10 for the workaround.

Will need to get to the bottom of this cc @f-bader

PR #621, opened 2025-01-14
Add support to post message to Team Channel via Webhook Uri parameter

wmacdonald10:
Wanted to be able to post a message to a Microsoft Teams Team Channel using the App Registration. 

Added **TeamChannelWebhookUri** parameter to **Invoke-Maester** and **Send-MtTeamsMessage**

To get the webhook Url, right-click on the channel in Teams and select 'Workflow'. Create a workflow using the 'Post to a channel when a webhook request is received' template. Use the value after complete

Here is an example of the post to a Team Channel

image


merill (approved):
THIS IS INCREDIBLE!!!

❤️ it. Thanks a lot @wesmacdonald

PR #629, opened 2025-01-17
Update for MS.AAD.5.4 - GroupConsent

weyCC81:
Ignore or skip Group Consent when not available

### DIFF
--- a/powershell/public/cisa/entra/Test-MtCisaAppGroupOwnerConsent.ps1
+++ b/powershell/public/cisa/entra/Test-MtCisaAppGroupOwnerConsent.ps1
@@ -23,6 +23,11 @@ function Test-MtCisaAppGroupOwnerConsent {
         return $null
     }
 
+    if ( $SettingsApiAvailable -notcontains 'EnableGroupSpecificConsent' ) {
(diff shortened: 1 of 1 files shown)


powershell/public/cisa/entra/Test-MtCisaAppGroupOwnerConsent.ps1
```diff
@@ -23,6 +23,11 @@ function Test-MtCisaAppGroupOwnerConsent {
         return $null
     }
 
+    if ( $SettingsApiAvailable -notcontains 'EnableGroupSpecificConsent' ) {
+        Add-MtTestResultDetail -SkippedBecause 'Custom' -SkippedCustomReason 'Settings value is not available. This may be due to the change that this API is no longer available for recent created tenants.'
```
SamErde: Should change "for recent created tenants" to "for recently created tenants." Is a more specific reason not available in the response?

SamErde (changes requested):

powershell/public/cisa/entra/Test-MtCisaAppGroupOwnerConsent.ps1
weyCC81: I copied this sentence from Eidsca test's, are we now changing this everywhere?

SamErde: I believe it should and was just looking got the best place to submit a PR.

PR #631, opened 2025-01-18
Update for CA Test - Minor Changes

weyCC81:
Add more text
Change value (Linux Container)
Generate a better output


merill:
Thanks good catch. I didn't realize the case sensitive issue on Linux containers. We might need to force case insensitive comparison going forward to avoid issues on Linux.

PR #632, opened 2025-01-18
Update for CA Test - Major Changes (Connect Sync)

weyCC81:
Warning: This change does change the behavior of the Test

**Do not check blocks for Legacy Auth**
In my opinion, blocks which are not targeting Modern Auth, should not be considered

**Ignore excludes for Synchronization Accounts**
As this service accounts are hardened by Microsoft, I believe we can skip them

</br>
_Reference: Entra ID Connect Sync / Entra Connect Sync_


merill:
Thanks. Agree with this. Although there is an upcoming change where we will need to revisit this entire test.

weyCC81:
> Thanks. Agree with this. Although there is an upcoming change where we will need to revisit this entire test.

Private Preview's for the Win! That's a good one.

PR #633, opened 2025-01-18
Update for CA Test - Major Changes (State: Enabled)

weyCC81:
Reviewed State for `Get-MtConditionalAccessPolicy`
- Add comment if not filtered for Enabled state
- Adjusted for simpler review of multiple tests


merill:
Thanks. This definitely helps to provide a more accurate picture, instead of passing tests when the CA policies are read only.

PR #634, opened 2025-01-21
Update Command Dependency Frequency Checks

SamErde:
This pull request replaces the existing `commandDependencyFrequency.ps1` script with an advanced function called `Get-CommandDependencyFrequency`. The new function provides enhancements that include rich object output and filtering options.

The function checks all public scripts and provide detailed output including command names, command usage counts, source modules, and file paths for each.


merill (approved):
Looks good!

Issue #626, opened 2025-01-17
seems get-mtuser count argument is not working as intended as object returns more than count limited


heetehub:
I seems the count break should be in foreach otherwise has no effect.

foreach ( $EntraIDRole in $EntraIDRoles ) {
                $TmpUsers = Invoke-MtGraphRequest -RelativeUri "directoryRoles/$($EntraIDRole.id)/members" -Select id, userPrincipalName, userType -OutputType Hashtable
                if ( $TmpUsers.ContainsKey('userType') ) {
                    Write-Verbose "Setting userType to Admin for $($TmpUsers.Count) users that are member of $($EntraIDRole.displayName)."
                    $TmpUsers | ForEach-Object {
                        $_.userType = "Admin"
                        $Users.Add($_) | Out-Null
                        # I put count check in loop
                         if ($Users.Count -ge $Count) {
                            Write-Verbose "Found $Count $UserType users."
                            break
                        }
                    }
                    **<#
                    if ($Users.Count -ge $Count) {
                        Write-Verbose "Found $Count $UserType users."
                        break
                    }
                    #>**

weyCC81:
Which Test or File are you referring to?

heetehub:
> Which Test or File are you referring to?

Get-MtUser.ps1 line 85, 120, 136,

heetehub:
Yes correct, as far as I see the issue presents itself only for admin and bg accounts.

merill:
I believe #641 has fixed this.

Please re-open if it still exists in new version.

PR #641, opened 2025-01-25
Adjust User Output in regards to Count

weyCC81:
Related to https://github.com/maester365/maester/issues/626#issuecomment-2602002239


merill:
Thanks for this fix @weyCC81

Issue #80, opened 2024-04-13
Azure DevOps - Run Maester Tests [Error]


dfiorentino:
Hi guys,

Do you have any ideas about this error ¿? I've created the YAML file as the documentation says and im using an appreg for auth.

![image](https://github.com/maester365/maester/assets/71448194/48c8580a-b029-41e7-8563-ee26f6f51b15)

dfiorentino:
I've added the ClientId and TenatId as a variable as the ClientSecret and it works!

PR #645, opened 2025-01-29
Add graph tier0 permission validation

HenrikPiecha:
Added test where app registrations are checked for having tier-0 graph permissions with a risk of having a direct or indirect path to Global Admin and full tenant takeover. 
This test was inspired by the Community Post in "Entra 🆔 News #80 → This week in Microsoft Entra" from "Emilien Socchi".

Further information: https://github.com/emiliensocchi/azure-tiering/tree/main/Microsoft%20Graph%20application%20permissions#tier-0

### DIFF
--- a/powershell/public/maester/entra/Test-MtHighRiskAppPermissions.md
+++ b/powershell/public/maester/entra/Test-MtHighRiskAppPermissions.md
@@ -1,0 +1,53 @@
+Ensure no graph application has permissions with a risk of having a direct or indirect path to Global Admin and full tenant takeover.
+
+This test if any application has tier-0 graph permissions with a risk of having a direct or indirect path to Global Admin and full tenant takeover.
+
+Following is a shortened copy from [Application permissions - Tier 0: Family of Global Admins](https://github.com/emiliensocchi/azure-tiering/tree/main/Microsoft%20Graph%20application%20permissions#tier-0), Date: 20.01.2025
+
+| Application permission | Path type | Known shortest path |
+| --- | --- | --- |
+| [AdministrativeUnit.ReadWrite.All](https://learn.microsoft.com/en-us/graph/permissions-reference#administrativeunitreadwriteall) | Indirect | When combined with other types of access allowing to reset user passwords, can remove a Global Admin from a [Restricted Management Administrative Unit (RMAU)](https://learn.microsoft.com/en-us/entra/identity/role-based-access-control/admin-units-restricted-management) and take it over. |
+| [Application.ReadWrite.All](https://learn.microsoft.com/en-us/graph/permissions-reference#applicationreadwriteall) | Indirect | Can impersonate any SP with more privileged application permissions granted for MS Graph, and impersonate it to escalate to Global Admin. |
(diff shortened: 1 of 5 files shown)


SamErde (changes requested):
This PR _should_ only contain changes related to the test description: 

> Added test where app registrations are checked for having tier-0 graph permissions with a risk of having a direct or indirect path to Global Admin and full tenant takeover.

This will make review, approval, and merging easier to safely manage.

HenrikPiecha:
Removed commits that are handeled in separate PR.

powershell/public/maester/entra/Test-MtHighRiskAppPermissions.md
```diff
@@ -0,0 +1,53 @@
+Ensure no graph application has permissions with a risk of having a direct or indirect path to Global Admin and full tenant takeover.
+
+This test if any application has tier-0 graph permissions with a risk of having a direct or indirect path to Global Admin and full tenant takeover.
```
SamErde: Grammar? "This test _checks_ if any application..."

SamErde (approved):
Nice test. I'm curious how the report could advise admins that might have an app with required permissions (grey area!), or if exceptions might ever be a possibility.

powershell/public/maester/entra/Test-MtHighRiskAppPermissions.md
HenrikPiecha: Has been adjusted.

HenrikPiecha:
Well... there are a lot of scenarios where at least permissions with an indirect path to a global administrator are required. Maester itself requires “RoleEligibilitySchedule.ReadWrite.Directory”, which also has a known indirect path to a global administrator.
This test should make the criticality of these permissions transparent to admins. They must assess whether an existing application can receive these privileges or not.

Should I extend the test-description with this information?

SamErde (approved):

SamErde:
> Well... there are a lot of scenarios where at least permissions with an indirect path to a global administrator are required. Maester itself requires “RoleEligibilitySchedule.ReadWrite.Directory”, which also has a known indirect path to a global administrator. This test should make the criticality of these permissions transparent to admins. They must assess whether an existing application can receive these privileges or not.
> 
> Should I extend the test-description with this information?

I do like the idea of putting that concept into the description, so admins don't get the impression that it is an all or nothing decision for 100% of real-world scenarios. 👍

HenrikPiecha:
Updated test description as discussed.

merill:
Thanks for adding this check @HenrikPiecha 

To be wholistic we need to query all the service prinicipals that have these privileged permissions.

We've held off on adding Applications related checks due to performance on large tenants (e.g. those who have 1/2 million+ service principals, etc..)

So for now I've made this test optional but it can be run with the `-Tag Full` switch.

Another limitation is that when we have large numbers, loading everything we have into memory  like we do today. 

For an upcoming release I'm working on a model where we do an upfront query and storing objects to a local db on disk to avoid memory limitations.

PR #646, opened 2025-01-29
Fix duplicate check of mfamigration

HenrikPiecha:
CISA Test "MS.AAD.3.5: The authentication methods SMS, Voice Call, and Email One-Time Passcode (OTP) SHALL be disabled." fails if MFA Migrations is not completet but the weak factors are disabled.
But CISA Test "MS.AAD.3.4: The Authentication Methods Manage Migration feature SHALL be set to Migration Complete." already checks if mfa migration is completed.

### DIFF
--- a/powershell/public/cisa/entra/Test-MtCisaWeakFactor.ps1
+++ b/powershell/public/cisa/entra/Test-MtCisaWeakFactor.ps1
@@ -35,15 +35,13 @@ function Test-MtCisaWeakFactor {
         "Email"
     )
 
-    $isMethodsMigrationComplete = Test-MtCisaMethodsMigration
-
     $result = Get-MtAuthenticationMethodPolicyConfig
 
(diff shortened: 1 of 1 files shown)


SamErde (changes requested):
This PR _should_ only contain changes related to the description:

> CISA Test "MS.AAD.3.5: The authentication methods SMS, Voice Call, and Email One-Time Passcode (OTP) SHALL be disabled." fails if MFA Migrations is not completet but the weak factors are disabled.
But CISA Test "MS.AAD.3.4: The Authentication Methods Manage Migration feature SHALL be set to Migration Complete." already checks if mfa migration is completed.

This will make review, approval, and merging easier to safely manage.

HenrikPiecha:
Removed commits that are handeled in separate PR.

merill (approved):
Looks good. Thanks @HenrikPiecha!

Issue #653, opened 2025-02-05
*Bug* Incorrect Typecasting of `RequestDurationInDays` in Eidsca CR04 Test Leading to Wrong Comparison Results


HCRitter:
**Description:**  
When testing the Eidsca CR04 in a tenant, the test returns a value of `5` days for `RequestDurationInDays`, which is clearly below the 30-day threshold. However, the test incorrectly reports that the duration is not 30 days or less. The issue is caused by an unnecessary and incorrect typecasting to a string, leading to a faulty comparison.

Failing code:
```PowerShell
$result = Invoke-MtGraphRequest -RelativeUri "policies/adminConsentRequestPolicy" -ApiVersion beta

[string]$tenantValue = $result.requestDurationInDays
$testResult = $tenantValue -le '30'
$tenantValueNotSet = $null -eq $tenantValue -and '30' -notlike '*$null*'
```

---

**Steps to Reproduce:**

1. Execute the following code snippet:
 ```powershell
   [string]$Tenantvalue = 5
   $TestResult = $TenantValue -le '30'
   $tenantValueNotSet = $null -eq $tenantValue -and '30' -notlike '*$null*'
```

2. Observe that the comparison is not behaving as expected.
3. Check the property type using:
```powershell
(Get-MgPolicyAdminConsentRequestPolicy).RequestDurationInDays.gettype()
```
The property is already an integer, which means that the cast to a string is unnecessary.

---

**Expected Behavior**

- The test should correctly compare the integer value without converting it to a string.
- A tenant value of `5` should be recognized as less than or equal to `30`.

---

**Actual Behavior**

- The string conversion leads to an incorrect comparison, resulting in the test always failing the condition.

---

**Suggested Fix**

- Remove the explicit string cast for RequestDurationInDays to ensure the comparison uses the native integer type.

---

**Source**

[Test-MtEidscaCR04](https://github.com/maester365/maester/blob/main/powershell/internal/eidsca/Test-MtEidscaCR04.ps1)

Cloud-Architekt:
@HCRitter: Thanks for the detailed feedback! I've worked on a fix to avoid explicit string cast for such kind of comparisons.
This issue will be resolved by PR: https://github.com/maester365/maester/pull/681

PR #654, opened 2025-02-05
Issue #653 fix: remove unnecessary string casting

HCRitter:
fix: remove unnecessary string casting and improve tenant value checks in admin consent policy test

Previously, the test for "policies/adminConsentRequestPolicy" incorrectly cast the RequestDurationInDays value to a string, resulting in faulty comparisons. This change removes the string cast, allowing for proper numeric comparison, and improves the handling of unset tenant values using [string]::IsNullOrEmpty() and a check for a zero value.

- Removed casting of RequestDurationInDays to string.
- Compared tenantValue directly against the integer 30.
- Updated tenantValueNotSet logic to reliably detect unset or default values.
- Ensured the testResultMarkdown messages remain consistent with the new logic.

### DIFF
--- a/powershell/internal/eidsca/Test-MtEidscaCR04.ps1
+++ b/powershell/internal/eidsca/Test-MtEidscaCR04.ps1
@@ -27,9 +27,9 @@ function Test-MtEidscaCR04 {
     }
     $result = Invoke-MtGraphRequest -RelativeUri "policies/adminConsentRequestPolicy" -ApiVersion beta
 
-    [string]$tenantValue = $result.requestDurationInDays
-    $testResult = $tenantValue -le '30'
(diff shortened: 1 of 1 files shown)


SamErde (approved):
✅👍

merill:
Good bug catch @HCRitter tx. 

The only catch is, this .ps1 file is auto-generated from the EIDSCA json on build and the change you made will be overwritten. We document the process here https://maester.dev/docs/contributing#updating-eidsca-tests-and-documentation

So to fix this issue, the file that needs to be edited is https://github.com/maester365/maester/blob/main/build/eidsca/Update-EidscaTests.ps1

Tagging @Cloud-Architekt

merill:
Closing as the code is auto generated and the checks are needed for other tests that check against default value.

PR #657, opened 2025-02-10
Detailed how-to create a CA policy for Device compliance

Appelcloud:
Detailed how-to guide for create a CA for Device compliance


Appelcloud:
@SamErde hah I see what you mean ;-) I could have added "CA Policy" instead 😎

SamErde:
> @SamErde hah I see what you mean ;-) I could have added "CA Policy" instead 😎

Haha, it's all good! I was very surprised while chatting with a couple of Microsoft product managers recently: when I referred to "CAPs" in a round-table conversation, they did not know what I meant (even though the topic of conversation was conditional access). "We've never referred to them as CAPs." Everyone else at the table seemed to get the reference. Interesting perspective. 🙂

merill:
This is a good update thanks!

If the guidance is for device compliance, we should remove the step on **Require multifactor authentication**

Appelcloud:
@merill you are rigth! "Require multifactor authentication" should be removed for this one. Good catch

PR #659, opened 2025-02-12
Add new test MT.1049

BakkerJan:
Add a new CA test: Checks for common misconfigurations in Conditional Access: both user risk and sign-in risk are configured in one policy.


merill (approved):
This is a super neat check. Thanks for adding it @BakkerJan !

PR #664, opened 2025-02-14
Changed the Graph endpoint from applications to servicePrincipals to …

HenrikPiecha:
## Test
Changed the Graph endpoint from applications to servicePrincipals to be wholistic. 
Made compare-adjustments in order to keep its functionallity as servicePrincipals have to be handeled differently than applications.

**Old:**
`
Invoke-MtGraphRequest -RelativeUri "applications"
`

**New:** 
`Invoke-MtGraphRequest -RelativeUri "servicePrincipals"`
per sp:
`Invoke-MgGraphRequest -Uri "https://graph.microsoft.com/v1.0/servicePrincipals/{Id}/appRoleAssignments" -Method GET`
`Invoke-MgGraphRequest -Uri "https://graph.microsoft.com/v1.0/servicePrincipals/{Id}/oauth2PermissionGrants" -Method GET`

Please note: `Invoke-MtGraphRequest` failed when trying to call the dedicated endpoints for querying the permissions assigned. Therefore I used the default `Invoke-MgGraphRequest` call.

### Application
Graph endpoint: servicePrincipals/{objectId}/appRoleAssignments
Only returns permissionId (appRoleId)
[List appRoleAssignments](https://learn.microsoft.com/en-us/graph/api/serviceprincipal-list-approleassignments?view=graph-rest-1.0&tabs=http)

### Delegated
Graph endpoint: servicePrincipals/{objectId}/oauth2PermissionGrants
Only returns the permissionName (Claim value)
[List oauth2PermissionGrants](https://learn.microsoft.com/en-us/graph/api/serviceprincipal-list-oauth2permissiongrants?view=graph-rest-1.0&tabs=http)

## Description
Added the direct URL of the application as a link in the report table.
Changed the result to also include the test result ($testResultMarkdown) and the report Table ($result).

## Test Call
Splitted the test into two separate tests to better distrinct between direct and indirect attack paths. This makes it easier to exclude, for example, the test of indirect attack paths to only include direct attack paths.
@merill I simply increased the previous test number by one.


weyCC81:
Please also publish some docs here (otherwise the numbers are used twice sometimes): https://maester.dev/docs/tests/MT.1050 & https://maester.dev/docs/tests/MT.1051

HenrikPiecha:
@weyCC81 Thank you for the feedback, didn't had that in mind. Added test descriptions for this PR via PR #666

PR #677, opened 2025-02-19
Detect Duo Etc Custom Controls

brianreidc7:
The check for all users are subject to MFA does not consider the nearly legacy "Custom Controls". If grantControls.customAuthenticationFactors is not "" then a custom control for external MFA is in place. This is typically the value "RequireDuoMfa" when it is used.


merill (approved):
This is a good catch. Thanks for helping make this better.

PR #679, opened 2025-02-20
Update Test-MtCaMfaForAllUsers.ps1

brianreidc7:
Updated to support legacy Custom Controls


merill (approved):
Thanks for the add @brianreidc7

PR #681, opened 2025-02-21
Added support for defining value type based on compare operator

Cloud-Architekt:


merill (approved):
Nice update Thomas. Thanks!

Issue #612, opened 2025-01-06
Invoke-Maester support for custom DNS resolvers


moorereason:
I need to bypass the default DNS resolvers in order to validate Exchange MX/SFP/DKIM records against an external resolver.  I would like the ability to do:

```
Invoke-Maester -DnsServerIpAddress 8.8.8.8
```

Is this possible today?

soulemike:
It isn't presented through to the core module today, though the Get-MailAuthenticationRecord command can support it.

It should be possible to add module variable and parameter to Invoke-Maester to set it. Then have the tests reference it when they call for the records.

In the interim though, the local copy of the tests could be modified to add that parameter when they call for the records.

moorereason:
From what I can tell, Invoke-Maester doesn't have any module variables or parameters that get passed down to subsequent commands like this.  If I'm missing it, please direct me to existing examples.

If not, what is the preferred design?  I'd be willing to work on solving this DNS resolver issue, but I'd like someone from the Maester team to sketch out a proposed API for adding such things to the project.

soulemike:
There are a couple options:

1. Add a module parameter and pass that into tests calling `Get-MailAuthenticationRecord` as a module variable.
2. Create custom tests for your environment that mirror the built in, but set the parameter on `Get-MailAuthenticationRecord` for your environment. Such as in https://github.com/maester365/maester/blob/main/powershell/public/cisa/exchange/Test-MtCisaDkim.ps1

Neither is wrong, both can be done concurrently too.

moorereason:
PR submitted to implement option 1.

Issue #40, opened 2024-03-01
Configuration file and function


f-bader:
Create a JSON that allows for environmental configurations
Create functions that will query the JSON and return the objects

fflaten:
Not `.psd1`? 🙂

lbrauns:
What configurations would be needed in there? I would work on this issue, my use case is to define the emergency access account(s) for the tenant. 

What format should be preferred? JSON would be a possibility for sure. Personally i have a preference for YAML. PSD1 would also be a suitable format. 

@merill maybe?

merill:
@FriedrichWeinmann's PSFramework has a really good configuration system. https://psframework.org/documentation/documents/psframework/configuration.html

It already supports exporting/importing to .psd1, clixml and json. https://psframework.org/documentation/documents/psframework/configuration/persistence-serialization.html

I'm thinking we can leverage that instead of re-inventing the wheel.

PR #684, opened 2025-02-21
Add DNS server IP address parameter to Invoke-Maester

moorereason:
To allow the user to pass a custom DNS server IP address down to the underlying CISA Exchange tests, add a cmdlet parameter named DnsServerIpAddress to Invoke-Maester.  The default value remains `1.1.1.1`.

To enable passing custom data down to the tests, I began using the $PesterConfiguration.Run.Container property.  The container will now hold the Path to the tests as well as the custom Data object. The DnsServerIpAddress parameter will be passed to all tests, but the parameter will be silently ignored by any tests without that parameter defined.

Updates: #612

### DIFF
--- a/powershell/public/Invoke-Maester.ps1
+++ b/powershell/public/Invoke-Maester.ps1
@@ -76,6 +76,11 @@ Invoke-Maester -PesterConfiguration $configuration
 ```
 Runs all the Pester tests in the EIDSCA folder.
 
+.EXAMPLE
+Invoke-Maester -DnsServerIpAddress '8.8.8.8'
+
+Run all the tests while using a custom server IP address for DNS lookups.
+
 .LINK
     https://maester.dev/docs/commands/Invoke-Maester
 #>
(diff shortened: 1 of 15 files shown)


soulemike:
Thanks for the PR! I am curious if there are pros/cons to passing this value through Pester versus the module itself.

moorereason:
I'm not a Maester or Pester expert (I'm new to both), so don't take my word for it.  After digging into how `Invoke-Maester` works, the command flow appears to be: `Invoke-Maester` → `Invoke-Pester` → `$CHECK.Tests.ps1` → `$CHECK.ps1`

The only way I could see to send parameters to a `$CHECK.ps1` module was to pass them through `Invoke-Pester`.  If there are more preferable ways to accomplish this, I'm open to it.

moorereason:
The major issue to this approach is that all tests would now share a parameter namespace from `Invoke-Maester`.  Whatever Data parameters we pass to `Invoke-Pester` will be passed to every test.  I could foresee more requests coming in to add additional parameters to be passed down to the tests.  We'd need to be careful how that's implemented.

merill:
@moorereason thanks for the PR.

This goes to the bigger feature around how to pass configuration parameters for various commands and tests (e.g. defining emergency access accounts, accepted number of global admins, etc)

We are tracking this in this feature and is something we've been wanting to implement for some time. 
https://github.com/maester365/maester/issues/40

This way we can use an extensible model to pass config parameters without having to update Invoke-Maester for each test.

Just haven't got around to implementing yet.

moorereason:
Closing since this isn't the right direction for the project.  I'll look into the PSFramework configurations, if I can find the time.

PR #689, opened 2025-02-22
Additional of EOP License Check (Fix#670)

soulemike:
Added license check for Exchange Online Protection and updated CISA tests to reference appropriately.


merill (approved):
Solid update. Thanks for the fix Mike!

PR #692, opened 2025-02-22
Update Convert-MtResultsToFlatObject.ps1

SamErde:
Add **ValueFromPipeline** to the **MaesterResults** parameter definition.

Fix end comment that still has "JSON" in the function name.


merill (approved):
Thanks Sam!

Issue #639, opened 2025-01-23
Open Issues with ORCA Tests


soulemike:
Ref #509 

Fails, but passes

- [ ] 103 - Misconfig
- [ ] 226 - Misconfig
- [ ] 227 - Misconfig
- [ ] 240 - Misconfig

No policy details

- [ ] 108 - No details
- [ ] 189 - No details
- [ ] 242 - Missing
- [ ] 243 - No details

Fails without MD

- [ ] 118 - Passed in ORCA
- [ ] 120 - Passed in ORCA

Passes without MD

- [ ] 143

Skipped, but passes

- [ ] 114 - Informational only in ORCA

-------

- [ ] There are a couple EXO cmdlets that are not set to cache in Get-ORCACollection due to parameters.
- [ ] Many of the tests don't return policy names, so it often appears like it is just checking the same setting multiple times. e.g., 100, 101, 102

moorereason:
I'm working on resolving these issues.  Hope to have some PRs after #698.

merill:
Awesome work @tdcthosc and @moorereason

PR #698, opened 2025-02-24
Simplify ORCA modules

moorereason:
Fix ORCA module generation to avoid duplicating the classes and enums to every Check module.  The root issue was that `orcaClass.ps1` needed to be renamed to `orcaClass.psm1`.

Issue originally identified [here](https://github.com/maester365/maester/issues/509#issuecomment-2613994736).

Related to: #639

Cc: @soulemike


moorereason:
I'd definitely prefer that you test it yourself, but it does work for me in a fresh PowerShell session:

```
PS> Import-Module .\powershell\Maester.psd1 -Force
PS> Connect-Maester -Service All -Privileged
PS> Invoke-Maester -Tag ORCA
```

I can confirm the nightmares.  It took a couple hours of troubleshooting and scouring the interwebs before I stumbled onto the fix.

merill (approved):
Thanks a lot @moorereason!! Looks good.

soulemike:
Got around to setting up a nightly build for the prerelease and tested this as part of it. Looks great, awesome work @moorereason!

PR #701, opened 2025-02-25
small tweaks to MT.1037 documentation

Appelcloud:
A few changes in #how to fix# section. Clickable link to Teams Admin center added, bold text and the word"toggle" changed to "switch" as I think it sounds better :-)


merill (approved):
Nice work. Thanks @Appelcloud

PR #703, opened 2025-02-26
Update Test-MtHighRiskAppPermissions.ps1

HenrikPiecha:
Noticed that the "resourceId" for MS Graph changed in different (dev-tenant vs. production) environments. Removed filter in PSObject generation.


merill (approved):
tx for the fix.

PR #713, opened 2025-02-28
Update Test-MtCisSharedMailboxSignIn.ps1

HenrikPiecha:
Object generation for validating which sharedmailbox users are enabled failed / always returned all mailboxes without account enabled state in productive environments.

Changed the generation of the mailbox objects.

### DIFF
--- a/powershell/public/cis/Test-MtCisSharedMailboxSignIn.ps1
+++ b/powershell/public/cis/Test-MtCisSharedMailboxSignIn.ps1
@@ -32,6 +32,11 @@ function Test-MtCisSharedMailboxSignIn {
     Write-Verbose "Getting all shared mailboxes"
     $sharedMailboxes = Get-MtExo -Request EXOMailbox | Where-Object { $_.RecipientTypeDetails -eq "SharedMailbox" }
 
+    if (($sharedMailboxes | Measure-Object).Count -eq 0) {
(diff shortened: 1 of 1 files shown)


NZLostboy:
I think this one needs a few more changes, in my testing now it doesn't list any shared mailboxes if it passes. If this is by design then I think the results need to be updated to reflect that. There are two shared mailboxes in the tenant I tested against.

Screenshot 2025-03-01 at 8 32 51 AM

NZLostboy:
Also the result logic for display is broken (I did this when I originally built it), if one mailbox fails, it shows them all failed.

Example, only one of these mailboxes has sign-ins enabled:
Screenshot 2025-03-01 at 8 44 27 AM

I realise you didn't make this error, but we should fix it while we are here.

HenrikPiecha:
@NZLostboy Added a Skip when there are no shared mailboxes in the tenant. 
Also changed the output of the result to only print the table when it failes and changed to table to include the displayname and the upn.
I think its not mandatory to add a "failed" result to each mailbox as its only printing the failed ones.

Let me know what you think. 

BR

NZLostboy:
> @NZLostboy Added a Skip when there are no shared mailboxes in the tenant. Also changed the output of the result to only print the table when it failes and changed to table to include the displayname and the upn. I think its not mandatory to add a "failed" result to each mailbox as its only printing the failed ones.
> 
> Let me know what you think.
> 
> BR

Approach is fine with me, looks great thank you.

merill:
Thanks for raising this and the fix!

PR #716, opened 2025-02-28
Changed Domain query to use API instead of cmdlet

HenrikPiecha:
Changed query of all domains to use the Graph API instead of the function that requires an additional cmdlet to be installed.


NZLostboy:
Looks good, nice update.

merill (approved):
Perfect

Issue #637, opened 2025-01-23
Removing filters/tags results in screen going black


jkerai1:
![Image](https://github.com/user-attachments/assets/af1ee970-02d5-4b93-8063-f95e41b29559)

![Image](https://github.com/user-attachments/assets/ad0f5348-6445-4ba8-8600-b643b4bf4f59)

zaulich:
**Observation:**

The issue with the blank HTML report seems to be related to tests categorized as "NotRun". The error props.Item.Tag.map is not a function suggests that Tag is sometimes a string or null instead of an array for these tests. This does not happen for Passed, Failed, or Skipped categories.

Just a observation :)

tdcthosc:
Should be resolved in #721

PR #721, opened 2025-03-04
Tags must always be an array

tdcthosc:
Otherwise props.Item.Tag.map() in ResultInfoDialog.jsx will result in blank screen


tdcthosc:
Should resolve #637

merill:
Thanks! I noticed this too when I had only one result.

Issue #710, opened 2025-02-28
MT.1052 - Device code auth flow test incorrect


biggles007:
The new MT.1052 test seems to have an issue. Microsoft are recommending that deviceCode is blocked by default, but the test is checking to ensure that device code is within at least one policy, but doesn't specifically check to see whether the policy is actually blocking.

Therefore, having a policy that grants access would fulfil the requirements of this test, but not actually secure anything.

https://github.com/maester365/maester/blob/c5353523b5c4bf258933807fc45d7da2daabb124/powershell/public/Test-MtCaDeviceCodeFlow.ps1#L29

tdcthosc:
Should be resolved in #722 as it will now check for various implementations of such conditional access

PR #722, opened 2025-03-04
Improve conditional access tests

tdcthosc:
Various improvements to multiple Conditional Access tests. Multiple commits for better change visibility and explanations on changes.

### DIFF
--- a/powershell/public/Test-MtCaDeviceCodeFlow.ps1
+++ b/powershell/public/Test-MtCaDeviceCodeFlow.ps1
@@ -30,7 +30,6 @@ function Test-MtCaDeviceCodeFlow {
             -and $policy.conditions.users.includeUsers -eq 'All' `
             -and $policy.conditions.clientAppTypes -eq 'all' `
             -and ($policy.grantcontrols.builtincontrols -contains 'block' `
-                -or ($policy.conditions.locations.includeLocations -eq 'All' -and $policy.grantcontrols.builtincontrols -contains 'block') `
                 -or ($policy.grantControls.builtInControls -contains 'compliantDevice' -or $policy.grantControls.builtInControls -contains 'domainJoinedDevice' )
             )
         ) {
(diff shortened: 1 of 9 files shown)


tdcthosc:
Should resolve #710

powershell/public/Test-MtCaDeviceCodeFlow.ps1
```diff
@@ -21,13 +21,19 @@ function Test-MtCaDeviceCodeFlow {
...
     $result = $false
 
     foreach ($policy in $policies) {
-        if ( $policy.conditions.authenticationFlows.transfermethods -eq "deviceCodeFlow")
-        {
+        if ($policy.conditions.authenticationFlows.transferMethods -contains 'deviceCodeFlow' `
+            -and $policy.conditions.users.includeUsers -eq 'All' `
+            -and $policy.conditions.clientAppTypes -eq 'all' `
+            -and ($policy.grantcontrols.builtincontrols -contains 'block' `
+                -or ($policy.conditions.locations.includeLocations -eq 'All' -and $policy.grantcontrols.builtincontrols -contains 'block') `
```
biggles007: There is no point in this line being in the criteria as the conditions are covered by the line above.

thomas-s-schmidt: I agree. However we do need to properly check the use of network locations. It should not be passed if "block" is chosen only for select networks. I have now implemented check for this. Please review.

merill:
@tdcthosc  this are some really good updates! Thanks a lot.

On the Device Code flow, since we are being strict about the policy, it will be nice to show the policies that target device code flow but don't meet our criteria.

This will help the user understand why we are failing the test even though they might have a device code block policy.

Thoughts?
 
CleanShot 2025-03-06 at 14 43 42

tdcthosc:
> @tdcthosc this are some really good updates! Thanks a lot.
> 
> On the Device Code flow, since we are being strict about the policy, it will be nice to show the policies that target device code flow but don't meet our criteria.
> 
> This will help the user understand why we are failing the test even though they might have a device code block policy.
> 
> Thoughts?

I agree and will get to work on that later today.

merill:
Thanks Thomas!

merill:
LGTM tx @tdcthosc

Issue #711, opened 2025-02-28
PIM Alert checks fail if there are no active alerts


biggles007:
If there are no active PIM alerts available then the MT.1029, MT.1030, MT.1031 and MT.1032 checks can't complete. You get the following error.

_RuntimeException: You cannot call a method on a null-valued expression.
at <ScriptBlock>, /home/vsts/.local/share/powershell/Modules/Maester/1.0.32/public/Test-MtPimAlertsExists.ps1:49_

https://github.com/maester365/maester/blob/c5353523b5c4bf258933807fc45d7da2daabb124/powershell/public/Test-MtPimAlertsExists.ps1#L49

The test is assuming there will be an alert with an active status, which isn't always the case. The test is filtering for active alerts of the require ID which returns a null object which the test then tries to process, hence the error above.

tdcthosc:
Should be resolved in #723

PR #723, opened 2025-03-04
Improve mt.1029-1032 tests

tdcthosc:
Links in alerts were HTML formatted, which needed to be converted to Markdown format.

If alerts did not have additional data, the outcome was not as expected.


tdcthosc:
Should resolve #711

merill (approved):
Nice. Thanks for fixing this.

Issue #670, opened 2025-02-18
Incorrect License Requirements for Exchange Online Tests


scs-ash:
A lot of the Exchange Online tests are being skipped due to licensing requirements not being met. The warning message states it needs Defender for Office 365 Plan 2, but I believe all of these functions are available in Defender for Office 365 Plan 1. The tenant I'm running this on has Microsoft 365 Business Premium, so I would expect these tests not to be skipped.

![Image](https://github.com/user-attachments/assets/95bee83e-91a5-4aef-8622-11519f352fe8)

![Image](https://github.com/user-attachments/assets/9e16285d-bd26-4a4e-9769-16d7f439511d)

merill:
Thanks for reporting.

Tagging @soulemike to review.

soulemike:
I would agree, these should all be achievable with EOP and not require MDO P1/P2. 

We can add an Eop product switch to https://github.com/maester365/maester/blob/b17961cd51d7f1d89e537f65c78ea56d36b23087/powershell/public/Get-MtLicenseInformation.ps1

It looks like EOP uses 326e2b78-9d27-42c9-8509-46c827743a17, I don't see any explicit inclusion for Commercial/Gov/Edu SKUs though. We can start with that and if there is issues then add like we have with the others.

@scs-ash do you feel comfortable making that PR, or would you prefer I do?

Test-MtCisaAttachmentFileType	MS.EXO.9.1v2
Test-MtCisaAttachmentFileType	MS.EXO.9.2v1
Test-MtCisaAttachmentFileType	MS.EXO.9.3v2
Test-MtCisaEmailFilterAlternative	MS.EXO.9.4v1
Test-MtCisaBlockExecutable	MS.EXO.9.5v1
Test-MtCisaAttachmentFilter	MS.EXO.10.1v1
Test-MtCisaMalwareAction	MS.EXO.10.2v1
Test-MtCisaMalwareZap	MS.EXO.10.3v1
Test-MtCisaImpersonation	MS.EXO.11.1v1
Test-MtCisaImpersonationTip	MS.EXO.11.2v1
Test-MtCisaMailboxIntelligence	MS.EXO.11.3v1

scs-ash:
> I would agree, these should all be achievable with EOP and not require MDO P1/P2.
> 
> We can add an Eop product switch to https://github.com/maester365/maester/blob/b17961cd51d7f1d89e537f65c78ea56d36b23087/powershell/public/Get-MtLicenseInformation.ps1
[...]

Issue #647, opened 2025-01-29
Migration of authentication methods can't be completed in newly created tenants as they are already migrated.


HenrikPiecha:
CISA Tests for validating the completion of migration for authentication methods fail in new tenants. 
Authentication methods migration can't be adjusted in new tenants as they are created in a "completed" state.

tdcthosc:
Should be resolved in #725

PR #725, opened 2025-03-04
Improve CISA tests

tdcthosc:
Various improvements to multiple CISA tests. Multiple commits for better change visibility and individual commit comments.


tdcthosc:
Should resolve #670

tdcthosc:
Should resolve #647

tdcthosc:
Should resolve some, if not all, issues mentioned in #639

tdcthosc:
Should temporarily resolve #612 as we should not rely on fixed IP adresses, but use the systems default DNS server, until we support environment variables

merill (approved):
Thanks @tdcthosc!

PR #726, opened 2025-03-04
Improve ORCA implementation

tdcthosc:
First of all, hats off to @soulemike for the ORCA implementation!

Various improvements and bugfixes to the Update-OrcaTests


tdcthosc:
Should resolve some, if not all, issues mentioned in #639

build/orca/Update-OrcaTests.ps1
```diff
@@ -207,33 +229,53 @@ function Test-$($content.func){
...
     `$obj = New-Object -TypeName $($content.func)
-    `$obj.Run(`$Collection)
-    `$testResult = (`$obj.Completed -and `$obj.Result -eq "Pass")
+    try { # Handle "SkipInReport" which has a continue statement that makes this function exit unexpectedly
+        `$obj.Run(`$Collection)
+    } catch {
+        Write-Error "An error occurred during $($content.func): `$(`$_.Exception.Message)"
+        throw
+    } finally {
+        if(`$obj.SkipInReport) {
```
moorereason, line 244: Can we completely skip generating any files for these checks instead of skipping here?  For example, skip earlier around line 136.  As in:

```powershell
$skip = [regex]::Match($content.content,"this.SkipInReport\s*=\s*[`$]True",$option)
if ($skip.Success) {
    continue
}
```

moorereason: Ha, I forgot that I submitted this approach as #705.

thomas-s-schmidt: I think we should still include them in Maester, but simply set it as skipped already in the "It" statement of "tests" file. I will get to work on that. Thank you for the suggestion!

But just to clarify my understanding of how this implementation of ORCA has been made by @soulemike, all of the Exchange/SCC commands will always be run regardless and cached. Each individual test simply runs the class on the cached results.

tdcthosc:
Marked as draft, as the following is to do:
- [ ] Improve Skip as suggested by @moorereason 
- [ ] Only add "Skip if not connected to Security and Compliance" if the test actually is a SCC test

build/orca/Update-OrcaTests.ps1
soulemike: There is definitely a consideration of how much logic we add to this build process versus adding to the upstream module.

In this instance this type of skip logic would help our module handle the caching more efficiently, which is more around our modules implementation, so I don't see any concerns customizing that here assuming it is dynamic.

Improving this to allow grouping of commands within Get-OrcaCollection to filter what calls to the EXO/SCC module are actually made would be awesome. So in summary I think you both are heading down the right path. Happy to brainstorm if it helps.

thomas-s-schmidt: I've done consideration and I think we should keep the way I've done it here, so that we still see it in Maester report as skipped. By putting the skip logic where I've done it, we also still get the markdown file to fill the test details section.

However, one negligible detail is that I've move the "skip" out of "finally" logic for better readability.

tdcthosc:
Create PR in upstream ORCA module to work with commit [d1ae7b4](https://github.com/maester365/maester/pull/726/commits/d1ae7b4eca9f70303b6e506f69a9ad26447464be)

PR: https://github.com/cammurray/orca/pull/326

build/orca/Update-OrcaTests.ps1
thomas-s-schmidt: May I set this conversation as **resolved**? 😄

merill: I think we can. Let's try this out and change based on feedback.

@tdcthosc does this mean we can merge this PR?

thomas-s-schmidt: Yes, I believe so. I've made some final adjustments:
- Moved the SkipInReport back into Finally block
- Included the improvements currently awaiting approval in [ORCA PR #326](https://github.com/cammurray/orca/pull/326)

tdcthosc:
I've done the merge with main, as there were conflicts on the SkipInReport approach from commit 7496541 that did not generate Check 143 like we do with this new approach.

merill:
Awesome. This is excellent progress!

PR #727, opened 2025-03-04
Add intune platform tests

HenrikPiecha:
Added Intune tests to check the platform for:

- Is automatic device clean-up configured
- Is default compliance policy set to "not compliant"

Added new api permission to web-page that is required to check Intune. These are required for the new tests to work. 
Added test description for new tests.


HenrikPiecha:
Removed commit "Test-MtCisPasswordExpiry.ps1" that was added in a different PR

HenrikPiecha:
@merill as we discussed added the intune related tests.

merill:
These are good tests. Thanks @HenrikPiecha 

I fixed some typos and removed the check for permission scope. We don't do it in other tests, also there can be other permissions that can be used (e.g. d..m..readwrite.all etc.. 

We'll let the error message explain this.

PR #755, opened 2025-03-12
ORCA cleanups

moorereason:
Three commits in this PR:

1. I was seeing some diff churn with the related links between upstream and my branch, so I sorted those to avoid whatever was causing that.
2. Return early and reduce complexity.
3. Fix a formatting issue in the markdown output and replace the complex string building one-liners, which are hard to review.  See commit message for more details.

Cc: @tdcthosc


moorereason:
BTW: the check failures were in `main` prior to this PR.  I'm not intending to fix those here unless requested.

merill (approved):
Thanks for the updates. LGTM

PR #756, opened 2025-03-12
CISA MS.EXO.4.1: fix aggregate reports bug and skip Microsoft domains

moorereason:
Fix aggregate reports showing the forensic reports value, skip initial or coexistence domains (such as onmicrosoft.com), and avoid doing DNS lookups on the same domain twice.


merill (approved):
LGTM

PR #758, opened 2025-03-14
Update MT.1023 description for P2 license compliance

MortenMynster:
Revise the description of the MT.1023 test to ensure it accurately reflects the requirement for users utilizing a P2 license.


merill (approved):
Good catch. Thanks!

Issue #759, opened 2025-03-16
Github Actions OperationStopped: UnAuthorized error


RkirkCBD:
For several weeks now my github action has been receiving an OperationStopped: UnAuthorized error.  I checked the federated credential sign in logs and signins have been successful.  Not sure where to look.  I'm running tests directly from the latest in the repository.

`True
{
  "Success": "true",
  "AzVersion": "12.1.0"
}
/usr/bin/pwsh -NoLogo -NoProfile -NonInteractive -Command $env:PSModulePath
/usr/share:/home/runner/.local/share/powershell/Modules:/usr/local/share/powershell/Modules:/opt/microsoft/powershell/7/Modules
Initializing Az Module Complete
Running Az PowerShell Script
/usr/bin/pwsh -NoLogo -NoProfile -NonInteractive -Command /home/runner/work/_temp/8c188c0c-f8a0-49c2-b87c-bc8a69245f85.ps1
Starting Maester tests
WARNING: You're using Az version 12.1.0. The latest version of Az is 13.3.0. Upgrade your Az modules using the following commands:
  Update-PSResource Az -WhatIf    -- Simulate updating your Az modules.
  Update-PSResource Az            -- Update your Az modules.
There will be breaking changes from 12.1.0 to 13.3.0. Open https://go.microsoft.com/fwlink/?linkid=[224](https://github.com/centerforbiologicaldiversity/maester-tests/actions/runs/13883350512/job/38844792039#step:2:245)1373 and check the details.
WARNING: Upcoming breaking changes in the cmdlet 'Get-AzAccessToken' :
The Token property of the output type will be changed from String to SecureString. Add the [-AsSecureString] switch to avoid the impact of this upcoming breaking change.
- The change is expected to take effect in Az version : '14.0.0'
- The change is expected to take effect in Az.Accounts version : '5.0.0'
Note : Go to https://aka.ms/azps-changewarnings for steps to suppress this breaking change warning, and other information on breaking changes in Azure PowerShell.
**OperationStopped: UnAuthorized**
Error: Error: The process '/usr/bin/pwsh' failed with exit code 1`

RkirkCBD:
Fixed - I had to add 

include_exchange: false 

to my main.yml file

Oppedijk:
Had the same issue, thanks for reporting.

Maybe reopen this one, as Exchange is optional, so it shouldn't fail on this?

Oppedijk:
[...]

PR #760, opened 2025-03-17
Patch-issue-759

Oppedijk:
Change the include_exchange to be optional (like it is in the documentation) so the Github action does not fail.
#759


merill (approved):
Thanks!

PR #762, opened 2025-03-17
Remove unnecessary connection dependency

tdcthosc:
Remove unnecessary connection dependency from Test-MtCisAttachmentFilterComprehensive


merill (approved):
Good catch. Thanks @tdcthosc

PR #767, opened 2025-03-18
Update Test-MtAppManagementPolicyEnabled.ps1

BakkerJan:
I deleted the Workload ID license check, as this feature is now free.


merill (approved):
Ah yes! Good catch. Thanks Jan!

PR #774, opened 2025-03-21
Move Test-ORCA143 in functions to export

tdcthosc:
Simple adjustment of placement


SamErde (approved):
Good idea to have them listed in order. It would be nice to have a build test that checks for missing FunctionsToExport and as a bonus, a function that sorts them. 🍰

merill (approved):
Tx. Looks good!

PR #775, opened 2025-03-21
Fix parameter handling for CSV/Excel file export

SamErde:
# CSV/Excel file export functionality

This pull request includes several improvements to the `Convert-MtResultsToFlatObject.ps1` script. The main changes involve better handling of the parameters to improve ease of use and avoid unnecessary errors being thrown.

## Add parameters

* Added the `ExportExcel` and `ExportCsv` parameters that were described in the comment-based help but not implemented.
* Reordered the export parameters to maintain consistent grouping of the CSV and Excel related parameters.

## Fix logic when checking for existing files with the same name

* Warn the user if they used the wrong extension in the file path.
* Only check if the user explicitly chose to export a file (this is where some errors were being thrown unnecessarily).

## General

* Improve some comments and help messages.
* Consistent use of spacing and line breaks in parameter definitions.

### DIFF
--- a/powershell/public/core/Convert-MtResultsToFlatObject.ps1
+++ b/powershell/public/core/Convert-MtResultsToFlatObject.ps1
@@ -146,6 +146,11 @@
 
         # Export the FlattenedResults list to a CSV if requested.
         if ($PSBoundParameters.ContainsKey('CsvFilePath')) {
+            if ($CsvFilePath -match '\.xlsx$') {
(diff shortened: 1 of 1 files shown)


SamErde:
Note: I realized in testing that the current structure produces no output if the function is run with no parameters. I'll address that in a separate PR.

merill (approved):
Looks good! Thanks Sam.

PR #778, opened 2025-03-24
Added Domain.Read.All permissions

SebastianClaesson:
The test; 
https://github.com/maester365/maester-tests/blob/63f75a86cee0fc15dae3f189e2eefe20a673652d/tests/cis/Test-MtCisPasswordExpiry.Tests.ps1

Requires Domain.Read.All

image


merill:
Thanks a lot for the PR @SebastianClaesson 

The root problem was the check that was being done on this test for scopes. It wasn't complete since we already request for Directory.Read.All so it should have worked with existing permissions.

I've removed the check from the test so it doesn't force adding a new permission (a45a022)

Will close this PR for now (PS this will only get pushed out in the new module).

PR #798, opened 2025-04-11
Added check for OnPremisesSyncEnabled

rafalfitt:
presence of OnPremisesImuutableId is not enough
for hybrid accounts OnPremisesSyncEnabled must be set


merill (approved):
LGTM thanks @rafalfitt

Issue #799, opened 2025-04-11
Prep for v1.1 Release Issues


soulemike:
- [x] AAD5.4 - API available, EIDSCA.CP01
- [x] Teams - Not connecting as part of service all
- [x] ORCA143 - Skip in report

FundMSG:
How to run v1.1 using Maester tes repositories in GITHUB. 
I am using install_prerelease: true  # Optional: Set to true if you want to use Measter Preview Build when running tests
But still it is running on v1.0.54.

SamErde:
I noticed some of these issues as well but was trying to get through a bit more testing before reporting. Glad(ish) it wasn't just me.

SamErde:
@soulemike, do you want to create sub-issues for each of these or dedicated issues?

soulemike:
> [@soulemike](https://github.com/soulemike), do you want to create sub-issues for each of these or dedicated issues? I'd be happy to take on the Teams connection one as a start.

Probably sub issue for any group of issues you'd want to tackle. Like the Teams connection is a good stand alone, but the MtDevice may be the same issue. I was mainly just making notes and then was going to follow on with commits later unless others grabbed them first.

soulemike:
> How to run v1.1 using Maester test repositories in GITHUB. I am using install_prerelease: true # Optional: Set to true if you want to use Measter Preview Build when running tests But still it is running on v1.0.54.

You may need to uninstall the module then install with prerelease switch. Should be like Install-Module Maester -AllowPrerelease

SamErde:
Hope this has been appropriately redundant: I just created separate issues 806-808 and submitted #805 to resolve all of the known issues with `Connect-Maester` and `Test-MtConnection` (particularly the Microsoft Teams connection).

FundMSG:
@soulemike I uninstalled and reinstalled the module, but still the same.

![Image](https://github.com/user-attachments/assets/172fbb44-c147-45a9-a0be-c5357aae0d59)

Partially, I am using GitHub to test it. I allowed the prerelease as "true," but the test count is the same. Any inputs?

![Image](https://github.com/user-attachments/assets/5d41cc5a-1713-47fb-b85d-d3dbd9e37e16)

soulemike:
@FundMSG .54 would have been a prerel version, should be up to .62 as of some additional merges recently. Make sure you are either cloning the tests folder from the repo or using the `Update-MaesterTests` command to get the latest tests too though. I am not sure if the run-maester-tests GHA does that or not.

PR #805, opened 2025-04-14
Fix MicrosoftTeams segments in Connect-Maester and Test-MtConnection

SamErde:
Update the `Test-MtConnection` function in the `powershell/public/core/Test-MtConnection.ps1` file.

* **Service Connection Checks:**
  * Add Microsoft Teams to the list of services checked by the function.
  * Include Teams when `All` is specified for the **Service** parameter.
  * Update the Azure connection check to include an API version in the `Invoke-AzRestMethod` call. This call fails when an API version is not specified.
  * Change the default value of the **Service** parameter from `'Graph'` to `'All'`.

Update the `Connect-Maester` function in the `powershell/public/Connect-Maester.ps1` file.

* **Service Connections:**
  * Connect to Microsoft Teams when the **Service** parameter is set to `All`.
  * Reduce output by piping the command to `Out-Null`.

Updates to both files:

* **Cleanup and Standardization:**
  * Move the function definition to the top of the script.
  * Use single quotes instead of double quotes whenever possible.
  * Changed to Pascal case for project consistency.

Supports #799 'Teams - Not connecting as part of service all.'

### DIFF
--- a/powershell/public/Connect-Maester.ps1
+++ b/powershell/public/Connect-Maester.ps1
@@ -1,4 +1,5 @@
-﻿<#
+﻿function Connect-Maester {
+   <#
 .SYNOPSIS
    Helper method to connect to Microsoft Graph using Connect-MgGraph with the required permission scopes as well as other services such as Azure and Exchange Online.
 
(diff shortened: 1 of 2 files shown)


merill (approved):
LGTM

PR #823, opened 2025-04-17
Fixed a typo in the docs

thomasvochten:


merill:
Thanks @thomasvochten Good catch!

I noticed a few other typos so I updated it to be more clearer.

Issue #825, opened 2025-04-17
Error on Invoke-Item $out.OutputHtmlFile with Invoke-Maester


blindzero:
Hi There,

first off, great project!

we set up a Bitbucket pipeline as [documented](https://maester.dev/docs/monitoring/bitbucket#create-bitbucket-repository-variables).
However, there are some "issues" that we would like to improve (e.g. documentation uses public-tests but not the one from the repository).

We got the issue, that within `Invoke-Maester` we get an error thrown:

```pwsh
VERBOSE: Graph: True
VERBOSE: Graph: True
VERBOSE: Invoking Graph: https://graph.microsoft.com/v1.0/organization
VERBOSE: True
VERBOSE: GET https://graph.microsoft.com/v1.0/organization with 0-byte payload
VERBOSE: received -byte response of content type application/json
VERBOSE: Graph: True
VERBOSE: Generating HTML report.
🔥 Maester test report generated at test-results/test-results.html
Invoke-Item: /root/.local/share/powershell/Modules/Maester/1.0.0/public/Invoke-Maester.ps1:334
Line |
 334 |                  Invoke-Item $out.OutputHtmlFile | Out-Null
     |                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | An error occurred trying to start process
     | '/opt/atlassian/pipelines/agent/build/test-results/test-results.html'
     | with working directory '/opt/atlassian/pipelines/agent/build'. No such
     | file or directory
Tests Passed ✅: 66, Failed ❌: 56, Skipped ⚫: 66
```

We looked into our pipeline and can't identify any issue, because test-results-files are created and saved as artifacts.
`$MaesterParameters` looks like

```pwsh
$MaesterParameters: 
Name                           Value
----                           -----
Path                           /opt/atlassian/pipelines/agent/build
OutputFolder                   test-results
PassThru                       True
OutputFolderFileName           test-results
PesterConfiguration            PesterConfiguration
```
`$PesterConfiguration` seems to be fine as well - as tests are performed and output of verbosity also prints out correctly.
Also so the `$results | Format-List Result, FailedCount, PassedCount, SkippedCount, TotalCount, TenantId, TenantName, CurrentVersion, LatestVersion` is working properly.

I brought it down to `Invoke-Maester.ps1` and `$NonInteractive`. However, not sure, if 
a) the documentation https://github.com/maester365/maester/blob/main/website/docs/monitoring/bitbucket.md needs to be changed and 
[...]

PR #826, opened 2025-04-17
#825 Fix Error with Invoke-Item on Invoke-Maester with Bitbucket pipeline

blindzero:
**Reason**
see also https://github.com/maester365/maester/issues/825 for this
* `$NonInteractive` parameter not in template file for Bitbucket pipeline
* result pipeline tries to open html file in browser

**Solution**
* adding `$NonInteractive = $true` to bitbucket pipeline template file


merill (approved):
Good catch! Thanks @blindzero

PR #839, opened 2025-04-19
Better skip handling to clarify why domains were skipped if they are …

soulemike:
…coexistence or initial domains for #838 rather than just not avialable.


merill (approved):
LGTM. Thanks @soulemike

PR #849, opened 2025-04-24
Updated recommended value for PR06

Cloud-Architekt:
Values equal or less to the recommended value of 10 will be accepted


merill (approved):
LG. Thanks @Cloud-Architekt

PR #851, opened 2025-04-24
Delete Slack typo.

michaelkennedy:
Remove reference to Slack (as this doc is referring to Teams).


merill (approved):
Good catch. Thanks!

PR #856, opened 2025-04-27
Fix typo in ResultInfoDialog.jsx

SamErde:


merill (approved):
Thanks @SamErde 

LOL, no one noticed this for so long.

PR #857, opened 2025-04-28
feat: Removed Github action dependency upon Az.Account

### LINKED ISSUE #853: Authentication without Az PowerShell
The Github Action sample currently uses [azure/powershell](https://github.com/maester365/maester/blob/f1bb5a14f07c562bcc1d632c4f4ec1e8b31a0ec0/action.yml#L99C1-L101C12) which in turn requires `enable-AzPSSession: true` in [azure/login@v2](https://github.com/maester365/maester/blob/f1bb5a14f07c562bcc1d632c4f4ec1e8b31a0ec0/action.yml#L75-L81)

This adds quite some time, just to do the login part.

You can get away with this code, as shown [here](https://github.com/svrooij/automation-issue-form/blob/5903e1a02117fd2d9432ca48860d492a5029c130/scripts/execute_action.ps1#L14-L21).

```powershell
  $token = $(az account get-access-token --resource https://graph.microsoft.com) | ConvertFrom-Json
  if ($null -eq $token) {
    Write-Host "Failed to get access token. Please check your Azure CLI authentication."
    exit 1
  }
  $graphToken = ConvertTo-SecureString -String $token.accessToken -AsPlainText -Force
  # Connect to Microsoft Graph with the token as secure string
  Connect-MgGraph -AccessToken $graphToken -ErrorAction Stop
```

If you can get away with not using the azure/powershell action I think you can save about 20 seconds on each Maester run. Which is not much but if you count the total amount of time that is spend running maester in all those tenants. That might be a big impact. 🌱☘️

svrooij:
Fixed #853


svrooij:
@merill I fully understand this error, but we might have to explicitly ignore it for this [file](https://github.com/maester365/maester/pull/857/files#diff-352e69c1e84d5a9d6f0d7afd65fc5d4d1bff57c9ed60760c6a4ad8ff4eb02f17R46). The as secure string is needed to be a drop-in replacement. Some sdks seem to expect the token as `SecureString` while others expect a regular `string`

action/Run-Maester.ps1
```diff
@@ -47,7 +47,7 @@ BEGIN {
     Write-Host 'Starting Maester tests'
 }
 PROCESS {
-    $graphToken = Get-AzAccessToken -ResourceUrl 'https://graph.microsoft.com' -AsSecureString
+    $graphToken = Get-MtAccessTokenUsingCli -ResourceUrl 'https://graph.microsoft.com' -AsSecureString
```
SamErde, line 50: Can you please update this to support the use of different environments (worldwide, GCC, GCC-High, etc)? I know that's not as simple of a change, and the current implementation has the global endpoint URL hard coded, so feel free to defer that improvement for a later PR.

svrooij: Are the application IDs or resources known for those requests?
As in all regions and then these resources. Graph, Teams, Exchange.

I recon we can make it accept a "cloud/region" and then use a table to get the correct tokens. If Azure cli supports that off-course

merill:
@svrooij  we might need to move the Get-MtAccessTokenUsingCli.ps1 to the /action folder where Run-Maester.ps1 is.

The /internal folder is meant for the PowerShell module and it's currently breaking the build sinc we have pester test rules for .ps1 files in the /internal folder.

merill:
Yes, will be useful in the core module. When we do need to use it and expose it as a cmdlet we can move it around. 

Trying to keep things simple for now.

merill (approved):
Awesome. Thanks @svrooij

action/Run-Maester.ps1
SamErde: I _think_ you would just need to account for the different API endpoints for each sovereign environment.
https://learn.microsoft.com/en-us/entra/identity-platform/authentication-national-cloud

SamErde:
> @svrooij we might need to move the Get-MtAccessTokenUsingCli.ps1 to the /action folder where Run-Maester.ps1 is.
> 
> The /internal folder is meant for the PowerShell module and it's currently breaking the build sinc we have pester test rules for .ps1 files in the /internal folder.

```output
[D:\a\maester\maester\powershell\tests\general\PSScriptAnalyzer.Tests.ps1 ► Invoking PSScriptAnalyzer against commandbase.Analyzing rule PSAvoidUsingConvertToSecureStringWithPlainText.All files should be compliant: TEST-PSScriptAnalyzer.Tests.xml#L0](https://github.com/maester365/maester/pull/857/files#annotation_34180496718)
Failed test found in:
  TEST-PSScriptAnalyzer.Tests.xml
  TEST-PSScriptAnalyzer.Tests.xml
Error:
  at $failedFiles | Should -BeNullOrEmpty, D:\a\maester\maester\powershell\tests\general\PSScriptAnalyzer.Tests.ps1:33
```

While it's nice to avoid exceptions if we can, we should be able to suppress the failing check for this one specific file if there's no other way around it. I believe the relevant suppression would be this:

```powershell
[CmdletBinding()]
# Suppress PSAvoidUsingConvertToSecureStringWithPlainText for the entire script
[Diagnostics.CodeAnalysis.SuppressMessage('PSAvoidUsingConvertToSecureStringWithPlainText', '')]
param ()
```

PR #861, opened 2025-04-28
MS.AAD2.2 to check notificationRecipients  and additionalRecipients

blindzero:
MS.AAD2.2 Test-MtCisaNotifyHighRisk.ps1 was only checking the `notificationRecipients` from `identityProtection/settings/notifications`. These only contain recipients based on Entra roles, not the additionally added recipients which are stored in `additionalRecipients`.

This change merges `$notficationRecipients = ($result.notificationRecipients + $result.additionalRecipients)` prior passing it to the filter to check if there is some for RiskUsersAlerts activated.


blindzero:
Tests fail because of missing connection variables- out of PR scope

merill (approved):
Looks good. Thanks for the fix @blindzero

Issue #865, opened 2025-04-29
Allow setting the TeamsChannelWebhookUri in Github action


svrooij:
It would be super cool if you could just specify the [TeamsChannelWebhookUri](https://maester.dev/docs/monitoring/teams) parameter from the github action.

Having these results directly into Teams feels like an awesome addition.

Will look into this next week, if that is ok with you guys?

PR #871, opened 2025-05-01
Added output variable and support for teams notifications

svrooij:
Fixed: Allow setting the TeamsChannelWebhookUri in Github action #865

### DIFF
--- a/action/Run-Maester.ps1
+++ b/action/Run-Maester.ps1
@@ -40,7 +40,16 @@ param (
     [bool]$DisableTelemetry = $false,
 
     [Parameter(Mandatory = $false, HelpMessage = 'Add test results to GitHub step summary')]
-    [bool]$GitHubStepSummary = $false
+    [bool]$GitHubStepSummary = $false,
+
+    [Parameter(Mandatory = $false, HelpMessage = 'Teams Webhook Uri to send test results to, see: https://maester.dev/docs/monitoring/teams')]
+    [string]$TeamsWebhookUri = $null,
+
+    [Parameter(Mandatory = $false, HelpMessage = 'Teams notification channel ID')]
+    [string]$TeamsChannelId = $null,
+
+    [Parameter(Mandatory = $false, HelpMessage = 'Teams notification teams ID')]
+    [string]$TeamsTeamId = $null
 )
 
 BEGIN {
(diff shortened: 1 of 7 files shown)


action/Run-Maester.ps1
```diff
@@ -40,7 +40,16 @@ param (
...
-    [bool]$GitHubStepSummary = $false
+    [bool]$GitHubStepSummary = $false,
+
+    [Parameter(Mandatory = $false, HelpMessage = 'Teams Webhook Uri to send test results to, see: https://maester.dev/docs/monitoring/teams')]
+    [string]$TeamsWebhookUri = $null,
+
+    [Parameter(Mandatory = $false, HelpMessage = 'Teams notification channel ID')]
+    [string]$TeamsChannelId = $null,
+
+    [Parameter(Mandatory = $false, HelpMessage = 'Teams notification teams ID')]
```
SamErde, line 51: Super nit pick: HelpMessage should be 'Teams notification **team** ID'. :)

svrooij: Good find!

action/Run-Maester.ps1
```diff
@@ -130,11 +139,24 @@ PROCESS {
         }
     }
 
+    if ([string]::IsNullOrWhiteSpace($TeamsChannelId) -eq $false -and [string]::IsNullOrWhiteSpace($TeamsTeamId) -eq $false) {
+        $MaesterParameters.Add( 'TeamChannelId', $TeamsChannelId )
+        $MaesterParameters.Add( 'TeamId', $TeamsTeamId )
+        Write-Host "Results will be sent to Teams Team Id: $TeamsTeamId"
+    }
+
```
SamErde, line 147: It could be useful here to output a warning if the parameter is used, but an empty or whitespace value is provided. (I think a warning that skips the Teams notification would be better than to fail the entire job.)

You could check this with a combination of `if ($PSBoundParameters.ContainsKey('TeamsChannelId') -or $PSBoundParameters.ContainsKey('TeamsTeamId'))` and `if ([string]::IsNullOrWhiteSpace($TeamsChannelId) -or [string]::IsNullOrWhiteSpace($TeamsTeamId))` to track the warning condition.

SamErde: You could also use parameter validation, but that would have the more direct effect of killing the automated action if valid channel and team ID parameters are not provided, so in this case, I like the approach of validating within the script.

SamErde (approved):
Nice enhancements!

merill:
I'll go ahead and merge this thanks @svrooij !!

@svrooij would be good to make the changes suggested by @SamErde when possible. Thanks!!

PR #880, opened 2025-05-02
fix: More efficient github summary writing

svrooij:


merill (approved):
Thanks for the improvements @svrooij

Issue #859, opened 2025-04-28
Test CIS.M365.2.1.1 will report as failed as Built-In policy is not editable


cstaubli:
Dear all

Test CIS.M365.2.1.1 will always reports as failed as Built-In policy does not follow CIS recommendations and is not editable. This concerns properties "AllowClickThrough" and "EnableForInternalSenders".

As a suggestion, wouldn't it be better to check active rules via Get-SafeLinksRule and then check the linked policies?

Like:

```powershell
$rules = Get-SafeLinksRule
foreach ($r in $rules)
{
    $policy = $policies | Where-Object {$_.Name -eq $r.SafeLinksPolicy}
    # check...
}
```
It's the first time I report an issue here. If desired and welcome, I can also create a merge request for this. Let me know.

Kind regards
Christian

cstaubli:
Similar issue exists in CIS.M365.2.1.2 and CIS.M365.2.1.7, policies named "default" and "Office365 AntiPhish Default" are checked instead of the (real) default ones

f-bader:
@cstaubli is the only differentiator the name? If so it might be a good idea to exclude them from the comparison?

merill:
Tagging @NZLostboy to this thread.

This is what CIS says

> Note: E5 Licensing includes a number of Built-in Protection policies. When auditing
policies note which policy you are viewing, and keep in mind CIS recommendations
often extend the Default or Build-in Policies provided by MS. In order to Pass the
highest priority policy must match all settings recommended.

I think the right check is to find the policy with the highest priority and check if it meets the criteria.

Thoughts @cstaubli @NZLostboy?

NZLostboy:
Yeah this looks like oversight on my part, @cstaubli  good spotting. I remember how I got here, I was struggling to understand the best way to deal with it, and no option seems perfect but you are correct the logic here isn't quite right. I was trying to consider a situation where someone might split the settings over two policies etc, but when you start factoring in who the policies apply to etc etc it gets over complicated.

@merill Good catch, I had actually missed the CIS note about priorities. Could be harsh if there is a sufficient policy with priority 1 but to be fair this is their recommendation, and we have been trying to follow it to the letter given the fact there are plenty of other tests which cover different scenarios.

I think we should do is use `Get-SafeLinksRule` to get the policy with priority 0, and then check the rules for that policy via `Get-SafeLinksPolicy`. The only extra step we need here I believe is to get the policy name for the policy with priority 0 first, and then just complete the same checks we already were.

[...]

PR #891, opened 2025-05-04
Merge potential fix for CIS 2.1.1

NZLostboy:
Potential fix for issues described in https://github.com/maester365/maester/issues/859

This test now grabs policy 0 (as per CIS recommendations) and checks that policy, rather than the default policy which cannot be changed.


merill (approved):
Thanks @NZLostboy this should be a good fix.

Issue #907, opened 2025-05-07
🪲 Maester tries opening report in Github


svrooij:
## Describe the bug

Using both the `latest` or the `preview` version of maester I see in the logs that it tries to open the html report. Which obviously fails on a Github Actions runner

## To Reproduce

1. Invoke-Maester ....

on github actions runner (Linux)

## Expected behavior

I Expect it not to do anything "interactive" in a CI pipeline

## Debug Output

<details>
<summary>Debug Output</summary>

```plain
....
🔥 Maester test report generated at test-results/test-results.html
/usr/bin/xdg-open: 882: www-browser: not found
/usr/bin/xdg-open: 882: links2: not found
/usr/bin/xdg-open: 882: elinks: not found
/usr/bin/xdg-open: 882: links: not found
/usr/bin/xdg-open: 882: lynx: not found
/usr/bin/xdg-open: 882: w3m: not found
xdg-open: no method available for opening '/home/runner/work/maester-action/maester-action/test-results/test-results.html'

Tests Passed ✅: 46, Failed ❌: 75, Skipped ⚫: 150
....
Dont like this banner as well... Maybe same issue?

███╗   ███╗ █████╗ ███████╗███████╗████████╗███████╗██████╗
████╗ ████║██╔══██╗██╔════╝██╔════╝╚══██╔══╝██╔════╝██╔══██╗
██╔████╔██║███████║█████╗  ███████╗   ██║   █████╗  ██████╔╝
██║╚██╔╝██║██╔══██║██╔══╝  ╚════██║   ██║   ██╔══╝  ██╔══██╗
██║ ╚═╝ ██║██║  ██║███████╗███████║   ██║   ███████╗██║  ██║
╚═╝     ╚═╝╚═╝  ╚═╝╚══════╝╚══════╝   ╚═╝   ╚══════╝╚═╝  ╚═╝ v1.0.83

To get started, install Maester tests and connect before running Maester:

md 'maester-tests'           
cd 'maester-tests'           
Install-MaesterTests         
Connect-Maester -Service All 
Invoke-Maester               

See https://maester.dev/ for more info.🔥
```

</details>

## Module Version

- `1.0.0`
- `1.0.83`

## Additional context

It seems it cannot detect correctly if it is running in interactive mode.

Ps. I don't like the banner that is shown when you import the module (when not in interactive move)

merill:
I've added the nonInteractive flag so it should prevent the welcome message etc from filling up the logs when running as a github action

PR #909, opened 2025-05-08
Suppress the welcome message if running in a non-interactive session.

SamErde:
@svrooij raised a valid point in #907 that the welcome message should not be shown in GitHub actions or other non-interactive sessions.

This pull request introduces a change to `powershell/Show-MaesterWelcome.ps1` that skips the welcome message in non-interactive environments.

The `$NonInteractive` variable to detects non-interactive environments based the `UserInteractive` environment variable, the use of a `NonInteractive` command-line argument for PowerShell, or specific CI environment variables (`GITHUB_ACTIONS` or `GITLAB_CI`).

### DIFF
--- a/powershell/Show-MaesterWelcome.ps1
+++ b/powershell/Show-MaesterWelcome.ps1
@@ -2,6 +2,13 @@
 [Diagnostics.CodeAnalysis.SuppressMessageAttribute('PSAvoidUsingWriteHost', '', Justification = 'Colors are beautiful')]
 param ()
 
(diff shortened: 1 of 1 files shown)


svrooij:
And can we suppress it in VSCode as well, this happened when I opened the maester folder in it?
I really don't like this new welcome message

VS Code

SamErde:
Good question. I'm happy to bring this to a discussion see what others think as well.

SamErde:
Created #911 🙂

merill:
Cheers. Tx @SamErde

Issue #873, opened 2025-05-01
Split of the Github Action


svrooij:
Currently this repo is the source of, a PowerShell module, the website and the github action.

Actions are meant to be documented with the readme in the root of the repo, the readme in the root describes more the module than the action. Meaning the action is not optimal documented.

I have several issues with this way and I want to suggest to make a repo just for the action.

Features and improvements:

- Documentation just for that action
- Explaining exactly what the action does
- Using the action does not need to fetch all the files from the repo just to load a single script
- Tagging the action version (thus not breaking existing configurations if the action changes)
- Suggesting everybody to use a tagged version (using just the `main` version is definitely not recommended, and they are actively telling everybody that you should only use a tagged version) 

I would add the following improvements to the action:

- Being able to set which version of the Maester PowerShell module would be used (version or `latest`)
- Being able to set the version of public tests to be used (tag or branch)

This would also make it easier to contribute to the action, since there is nothing else in the new repo.

## Migrate

I'm very happy with the latest version of the action that I [built](https://github.com/svrooij/maester-action). It only took me 40 tries 😏

How do we continue from this? My idea is:

- [X] Create new [repo](https://github.com/maester365/maester-action) `maester365/maester-action`
- [X] Commit the code there (1 commit, not the mess I made)
- [x] Publish the action
- [ ] Update docs to point to the new action #910 
- [ ] Update old action to Output a warning to the console and an error in the run (with set-error command) and then call the new action. #910 
- [ ] Setup **issue templates** in both repos to point to the place for the issues.

SamErde:
Great explanation, @svrooij!

merill:
I think it makes sense. @f-bader thoughts?

f-bader:
Hi @svrooij definitely a good idea.

We have to solve two issues to move forward

* Name of the new repo \
   The main reason I choose the current implementation was to use maester365/maester as the name \
   maester365/maester-action would be an obvious name, Not really sleek but it works.
* How to migrate already existing workflows (manual action be the user)
  * We should add a warning message to the current action that explains the migration steps
  * We should actively post this on social media
  * We must keep the current action to not break stuff

I would love to hear you input on this

svrooij:
@f-bader and @merill I had the same idea. But it not an easy job.

I started here https://github.com/svrooij/maester-action and trying some stuff out. The plan is I create it here, and then you guys that a copy of the code and make a new repo in the correct org. Without all the 100 tries of me messing up.

[...]

PR #910, opened 2025-05-08
Replace action with warning and calling new action

svrooij:
Related to #873

1. Depreciation and migration page
2. Workflow runs show a notice (not a warning) pointing to the migration guide
3. This page is also written to the workflow summary
4. Adjusted `Readme` to point to migration


SamErde (commented):
Would it be helpful (and less scary) to admins to start this transition with a "deprecation" message that informs them of the new action source?

svrooij:
@SamErde Yes that will be better. Let me change that

svrooij:
@SamErde what do you think of this pr now?

1. Depreciation and migration page
2. Workflow runs show a notice (not a warning) pointing to the migration guide
3. This page is also written to the workflow summary
4. Adjusted `Readme` to point to migration

SamErde:
> Seeing how many commits I needed I would say this needs to be **squashed** into a single commit @merill

I should probably make use of squash commits more often...

README.md
```diff
@@ -78,50 +79,49 @@ Update-MaesterTests
 
 ## Use as GitHub action
 
-Maester is also published to the [GitHub marketplace](https://github.com/marketplace/actions/maester-action) and can be used directly in any GitHub workflow.
+Maester is also published to the [GitHub marketplace](https://github.com/marketplace/actions/run-maester) and can be used directly in any GitHub workflow. Because it is specially built for GitHub it nicely integrates with all the features of GitHub Actions, like uploading artifacts and writing a summary to the workflow run.
```
SamErde: Nit-pick: The editor in me suggests adding a comma after "built for GitHub[,]...". ✏️

README.md
```diff
@@ -78,50 +79,49 @@ Update-MaesterTests
...
 
-Just provide the required client and tenant id. For more details please refer to the [docs](https://maester.dev/docs/monitoring/github/).
+For more details please refer to the [docs](https://maester.dev/docs/monitoring/github/) or the [action repository](https://github.com/maester365/maester-action).
```
SamErde: Nit-pick: Sorry, the Comma Cop is calling again. "For more details[,] please refer..." ✏️

README.md
```diff
@@ -78,50 +79,49 @@ Update-MaesterTests
...
         tenant_id: ${{ secrets.AZURE_TENANT_ID }}
         include_public_tests: true # Optional
+        include_private_tests: true # Optional
         include_exchange: false # Optional
         include_teams: false # Optional
         pester_verbosity: None # Optional - 'None', 'Normal', 'Detailed', 'Diagnostic'
-        mail_recipients: '' # optional a list of email addresses to send the report to
-        mail_userid: '' # optional the user id to use for sending the email
-        include_tags: '' # optional a list of tags to include in the test run
-        exclude_tags: '' # optional a list of tags to exclude in the test run
```
SamErde, line 120: When I am working with an unfamiliar action configuration, I find it helpful when optional parameters are included but commented out. For example: 

```yaml
        # mail_recipients: '' # optional a list of email addresses to send the report to
        # mail_userid: '' # optional the user id to use for sending the email
        # include_tags: '' # optional a list of tags to include in the test run
        # exclude_tags: '' # optional a list of tags to exclude in the test run
```

SamErde (changes requested):
Looks awesome! I only had to nit-picks and one minor suggestion for the `action.yml` itself. Otherwise, I have no real objections, concerns, or complaints at all with the PR. Really great update!

README.md
svrooij: And that is why I want to direct everybody to the action documentation. Eventually I think this part has to go from the main readme, opinions?

SamErde: I'm good with that. Directing people to the documentation instead of leaving potential configuration is probably better because those commented optional configs could change in the future. Let's stick with your deletion of those lines.

merill:
There is one change I'm planning to make before we merge this in.

Right now the action pulls the tests from the repo, however there can be a mismatch between the Maester module version the user is running vs the tests. Especially now that the action can specify the version you want to run (preview vs production).

So instead of using that version we can update the action to run the `Update-MaesterTests` on the public tests folder. This way the tests will always be in sync with version of the tests embedded in the module.

Let me do that change in the new action and we'll test and then merge this in.

SamErde:
So the Maester action can be versioned independently from the module, but will still use the latest tests. 

S things develop, it could be interesting to revisit the concept of version management for each of the tests—or at least for the suites of tests. Then all three could be maintained individually.

svrooij:
I’ll resolve the feedback Monday.
Great point to change the action to use Update-tests instead of checking out the repo, there will be mismatches otherwise

merill:
@svrooij I've made the change to use the tests from the PowerShell module and simplified the other steps too.

Feel free to test out. Once your changes are in I'll merge this in.

merill:
This has been an awesome effort. Thanks a lot @svrooij for starting this initiative and @SamErde for all the feedback. All while @f-bader is living the good life in a beach somewhere 😂

PR #912, opened 2025-05-08
MT1057 App registrations should not have secrets defined

svrooij:
This pr needs to be squashed as we don’t want 4 commits just to add a single test


merill:
I added the -Full tag, since this test can take a long time in enterprise tenants with 100k> apps.

PR #928, opened 2025-05-15
fix tdcthosc author

tdcthosc:
I noticed your fix @merill – my apologies of not paying attention for duplicate 🙂 

I've made my nickname more unique and updated it on the blog post as well.


merill (approved):
Good catch. Thanks @tdcthosc !

PR #940, opened 2025-05-20
Remove welcome message from ScriptsToProcess

SamErde:
## Changes to Manifest

- Remove the welcome message from ScriptsToProcess. Instructions may be re-added at a later time in an interactive mode.
- Sort FunctionsToImport alphabetically.

## Show-MaesterWelcome

The `Show-MaesterWelcome` function is no longer actively used but was still not deleted.

- Update non-interactive checks to detect more CICD and containerized environments.
- Fix improper use of `-ne $null` on the right side of expressions.


merill (approved):
Cheers. Thanks Sam!

PR #950, opened 2025-05-22
fix #949: now a single dialog opens when you click a result.

### LINKED ISSUE #949: Report opens multiple dialogs
## Describe the bug
When clicking on a test it opens multiple dialogs after it has been refactored.

## To Reproduce

1. Open a report
2. Click on a test
3. Scroll to the bottom and you can see that multiple dialogs are open. It appears in the dev tools that it opens 3 dialogs:

Image

The background also appears darker than usual.

## Expected behavior
It should only open 1 dialog.

## Additional context
I'd suggest instead of using "item.Id || item.Name" several places for indexing, simply use the index value already present in the json for each test.

GuidoBaijense:
The `ResultInfoDialog` was opening multiple times because each row in the `TestResultsTable` was designed to have several clickable elements (ID, Title, and an Info button) that could each trigger a dialog for the *same* test result item.

Initially, the logic to determine if a dialog should be open was based on a shared identifier for the row's item (like `item.Id` or `item.Index`). When you clicked one element, the condition to open the dialog became true for *all* `ResultInfoDialog` instances associated with that same item's `Index`, causing all of them to appear.

The fix involved changing the `activeDialog` state in TestResultsTable.jsx to store a more specific identifier (a unique `dialogId` string like `dialog-${item.Index}-id`, `dialog-${item.Index}-title`, or `dialog-${item.Index}-button`). Each `ResultInfoDialog` instance was then given its own unique `dialogId`. The `isOpen` prop for each dialog was then determined by checking if the global `activeDialog` state matched its specific `dialogId`. This ensured that only the one specific dialog instance that was clicked (or navigated to) would open.

### DIFF
--- a/powershell/assets/ReportTemplate.html
+++ b/powershell/assets/ReportTemplate.html
@@ -61865,32 +61865,32 @@ function remarkGfm(options) {
   toMarkdownExtensions.push(gfmToMarkdown(settings));
 }
 function ResultInfoDialog(props) {
-  const itemId = props.Item.Id || props.Item.Name;
+  const itemIndex = props.Item.Index;
   const [isOpen, setIsOpen] = React.useState(false);
   const openInNewTab = (url) => {
     window.open(url, "_blank", "noreferrer");
   };
   reactExports.useEffect(() => {
-    if (props.activeDialog === itemId) {
+    if (props.isOpen) {
       setIsOpen(true);
-    } else if (isOpen && props.activeDialog !== itemId) {
+    } else if (isOpen && !props.isOpen) {
       setIsOpen(false);
     }
-  }, [props.activeDialog, itemId, isOpen]);
+  }, [props.isOpen, isOpen]);
   reactExports.useEffect(() => {
-  }, [isOpen, itemId]);
+  }, [isOpen, itemIndex]);
   reactExports.useEffect(() => {
     const handleKeyboard = (event) => {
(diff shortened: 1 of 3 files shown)


moorereason:
I would advocate for keeping the comments in the JSX files.

GuidoBaijense:
> Only one dialog opens now, however there is still quite the performance hit when the report contains several hundred tests. Do you also experience a delay when opening and closing dialogs, if the report contains hundreds of tests? I also see msedge.exe "spike" on a single core.
> 
> Before refactoring it takes approximately 80-100ms, and sometimes down to 40ms, to open a dialog.
> 
> This should be taken into considerations if you can reproduce it.
@tdcthosc 
O Wow. That's a huge difference and should indeed be taken in consideration. Is this all due to the version where I added pagination with arrow-key or do the added severity labels also have an impact?
What did you use to track performace?



> I would advocate for keeping the comments in the JSX files.

@moorereason I think my agent was a bit to aggresive when I asked it to "remove the comments I just added in the last commit". A comment that says: "function for left arrow key" while the function is called "leftarrowkey()" is not necessary, while others might.

I changed the PR to draft as I don't have time this week to address your valid points. Feel free to contribute of course.

merill (approved):
Love it. This definitely makes it more snappy and the code is cleaner too. 

Thanks for pushing this through!

PR #952, opened 2025-05-23
docs: add missing colon in Update-MgPolicyAuthorizationPolicy command value

MarcoJanse:
Update the EIDSCA.AP05.md page to add missing colon in `Update-MgPolicyAuthorizationPolicy` command parameter value


svrooij:
@MarcoJanse I never knew you would need a colon to set a Boolean parameter in PowerShell. Are you sure the old code did not work?

merill:
Yes for switch parameters the : is needed. If not it will be treated as true.

E.g. Update-MgUser -Verbose

merill (approved):
LG. Thanks @MarcoJanse

PR #956, opened 2025-05-27
fix: store tenantValue variable as integer for proper test outcome with comparison operator

MarcoJanse:
The [`Test-MtEidscaPR05.ps1`](https://github.com/maester365/maester/blob/main/powershell/internal/eidsca/Test-MtEidscaPR05.ps1) is storing the Password `LockoutDurationInSeconds` value as a string.

This resulted in incorrect test results as the tenant value was tested with the greaterThanOrEqual (`-ge`) comparison operator as a string instead of numeric value.

For example, in my tenant I had a `LockoutDurationInSeconds` value of `300`

When comparing this with the current Maester function module 1.1.0, I get the following results

> Test result ❗ 
>
> Your tenant is configured as 300.
>
> The recommended value is greater than or equal to '60' for settings

This is because in string comparison a value that starts with a 3 will always be lower than 6, so 300 is lower than 60 in a string comparison.

Results with on my tenant with the old function

```powershell
> (Test-MtEidscaPR05) -ge 60
False
```

With the modified function

```powershell
> (Test-MtEidscaPR05) -ge 60
True
```

## Environment

Tested module version: Maester 1.1.0

### DIFF
--- a/powershell/internal/eidsca/Test-MtEidscaPR05.ps1
+++ b/powershell/internal/eidsca/Test-MtEidscaPR05.ps1
@@ -27,7 +27,7 @@ function Test-MtEidscaPR05 {
     }
     $result = Invoke-MtGraphRequest -RelativeUri "settings" -ApiVersion beta
 
-    [string]$tenantValue = $result.values | where-object name -eq 'LockoutDurationInSeconds' | select-object -expand value
+    [int]$tenantValue = $result.values | where-object name -eq 'LockoutDurationInSeconds' | select-object -expand value
     $testResult = $tenantValue -ge '60'
     $tenantValueNotSet = ($null -eq $tenantValue -or $tenantValue -eq "") -and '60' -notlike '*$null*'
 


merill:
Good catch. Thanks @MarcoJanse for the PR!

Unfortunately this cmdlet is autogenerated based on EIDSCA json.

See details here: https://maester.dev/docs/contributing#updating-eidsca-tests-and-documentation

So the right fix will be to update /build/eidsca/Update-EidscaTests.ps1.

@Cloud-Architekt Should we do the casting to int or number whenever we use the lt, gt and other comparison operators? Thoughts?

MarcoJanse:
Hi @merill. Thank you for clarifying how that works. 

I have looked at the [Update-EidscaTests.ps1](https://github.com/maester365/maester/blob/main/build/eidsca/Update-EidscaTests.ps1), but I don't quite understand where the `[string]$TenantValue` is coming from and - as it's used in most Eidsca-tests - what will be the impact on the other tests. 🤔

merill:
Yes, there will be some impact to other tests, that's why we need to be careful. I think for anything other than eq we parse to integer. 

@Cloud-Architekt do you think that will work? Ideally we should add a type option to the EIDSCA json for instances where this default behavior doesn't work.

merill:
Tagging @Cloud-Architekt

MarcoJanse:
Thank you @Cloud-Architekt! I will close this pull request.

PR #963, opened 2025-06-04
handle multiple CAP covering LegacyAuth

andremieth:
The current Where-Object expects that both AppTypes ("other" and "exchangeActiveSync") are set in the same Conditional Access Policy (CAP). With the new logic, it will also handle situations where the AppTypes are split across more than one CAP.

### DIFF
--- a/powershell/public/cisa/entra/Test-MtCisaBlockLegacyAuth.ps1
+++ b/powershell/public/cisa/entra/Test-MtCisaBlockLegacyAuth.ps1
@@ -31,12 +31,22 @@ function Test-MtCisaBlockLegacyAuth {
 
     $result = Get-MtConditionalAccessPolicy | Where-Object { $_.state -eq "enabled" }
 
-    $blockPolicies = $result | Where-Object {`
-            $_.grantControls.builtInControls -contains "block" -and `
-            $_.conditions.clientAppTypes -contains "exchangeActiveSync" -and `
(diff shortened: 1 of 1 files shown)


powershell/public/cisa/entra/Test-MtCisaBlockLegacyAuth.ps1
```diff
@@ -31,12 +31,22 @@ function Test-MtCisaBlockLegacyAuth {
 
     $result = Get-MtConditionalAccessPolicy | Where-Object { $_.state -eq "enabled" }
 
-    $blockPolicies = $result | Where-Object {`
-            $_.grantControls.builtInControls -contains "block" -and `
-            $_.conditions.clientAppTypes -contains "exchangeActiveSync" -and `
-            $_.conditions.clientAppTypes -contains "other" -and `
-            $_.conditions.users.includeUsers -contains "All" }
+    $blockOther = $result | Where-Object {
+        $_.grantControls.builtInControls -contains "block" -and `
```
moorereason: While you're here, remove all backticks from this file.  They're unnecessary.  See [Get-PowerShellBlog](https://get-powershellblog.blogspot.com/2017/07/bye-bye-backtick-natural-line.html#logic).

Also, line 55 in your new file may need to be updated to something like "Your tenant lacks sufficient conditional access policies that block legacy authentication."  A similar issue exists on line 11.

PR #964, opened 2025-06-04
Update azure-devops.md - Get-AzAccessToken is now a secure string

mantis:
Get-AzAccessToken now returns as a Securestring, not secure string so no need to convert - see https://learn.microsoft.com/en-us/powershell/module/az.accounts/get-azaccesstoken?view=azps-14.0.0


merill (approved):
Perfect. Thanks for the PR @mantis

PR #965, opened 2025-06-05
Updated logic in Test-MtCaReferencedGroupsExist to fix missing output

Cloud-Architekt:
Missing try catch on `Invoke-MtGraphRequest` to look for deleted or invalid group object has led into failure without adding affected group object to list for test output.


merill (approved):
Looks good

PR #972, opened 2025-06-09
Updated CIS 2.1.2 Check

NZLostboy:
Updated check to use the IsDefault field rather than matching the name.


merill (approved):
Looks good. Thanks!

PR #974, opened 2025-06-16
fix test MT1024

svrooij:
It had the tenant id in the test id


SamErde:
_Very_ nice job fixing MT.1024 and MT.1039. The one interesting thing that I noticed is that the report shows `MT.1024.{1 - 9}` and then two occurences of `MT.102.unusedEnterpriseApps`. It would be cool if they could all be numbered or all labeled.

PR #983, opened 2025-06-23
Fix manifest required module version

SamErde:
https://github.com/maester365/maester/actions/runs/15813200285/job/44567504608#step:5:1

I mistakenly used Minimum Version instead of Module Version

This fixes the manifest and the publish preview workflow.


merill (approved):
Cheers. Tx was just about to look into this.

PR #988, opened 2025-06-24
fix: Test-MtCisaRequireActivationApproval.ps1 result markdown to represent test.

ThorNicolai:
Adjusted the test result markdown for "Test-MtCisaRequireActivationApproval.ps1".

Was not in line with the test, most likely the markdown was copied from a different test.


merill (approved):
Good catch. Thanks a lot @ThorNicolai

PR #989, opened 2025-06-24
Update export-results.md -fix parameter name

hbrowne22:
The parameter name is -InputObject rather than -Results


merill (approved):
Good catch. Thanks!

PR #995, opened 2025-06-26
Dynamic drift tests anyone?

svrooij:
I would say this test is just an experiment to see if this can be useful to anyone.

Any subfolders added to the new `drift` folder will be dynamically discovered as a drift test suite. Checking on a property-by-property level is the files `baseline.json` and `current.json` match.
The baseline has to be checked in, and then the current has to be generated right before the test run.

### DIFF
--- a/powershell/public/Invoke-Maester.ps1
+++ b/powershell/public/Invoke-Maester.ps1
@@ -183,7 +183,11 @@ function Invoke-Maester {
 
         # Do not show the Maester logo.
         [Parameter(HelpMessage = 'Do not show the logo when starting Maester.')]
-        [switch] $NoLogo
+        [switch] $NoLogo,
+
+        # Drift tests
+        [Parameter(HelpMessage = 'Specify drift root directory, see https://maester.dev/docs/tests/MT.1060')]
+        [string] $DriftRoot
     )
 
     function GetDefaultFileName() {
@@ -363,6 +367,21 @@ function Invoke-Maester {
         Write-Host "💫 Update-MaesterTests" -NoNewline -ForegroundColor Green
         Write-Host " → Get the latest tests built by the Maester team and community." -ForegroundColor Yellow
         return
+    }
+
+    # If DriftRoot is specified, set the environment variable for drift tests
+    if ($DriftRoot) {
+        $DriftRoot = (Resolve-Path -Path $DriftRoot -ErrorAction SilentlyContinue).Path
+        if (-not (Test-Path -Path $DriftRoot)) {
+            Write-Warning "❌ The specified drift root directory '$DriftRoot' does not exist." -ForegroundColor Red
+
+        } else {
+            Set-Item -Path Env:\MEASTER_FOLDER_DRIFT -Value $DriftRoot
+            Write-Verbose "🧪 Drift root directory set to: $DriftRoot"
+        }
+    } else {
+        # Default drift root directory
+        # Set-Item -Path Env:\MEASTER_FOLDER_DRIFT -Value $(Join-Path -Path (Get-Location) -ChildPath "drift")
     }
 
     $maesterResults = $null
--- a/tests/Maester/Drift/MT1060Drift.tests.ps1
+++ b/tests/Maester/Drift/MT1060Drift.tests.ps1
@@ -31,17 +31,18 @@ BeforeAll {
 # This allows the user to "define" drift tests by creating folders in the "drift" directory.
 BeforeDiscovery {
 
-    # Ensure the Compare-MtJsonObject function is available
-    if (-not (Get-Command -Name Compare-MtJsonObject -ErrorAction SilentlyContinue)) {
-        Write-Warning "Compare-MtJsonObject function missing, not the right version of Maester?"
+    # Get root directory for drift tests
+    # This assumes the drift tests are located in a folder named "drift" at the root of the current location
+    # $driftRoot = Join-Path -Path $(Get-Location) -ChildPath "drift"
+    $driftRoot = $env:MEASTER_FOLDER_DRIFT
+    # Ensure the drift root directory exists
+    if ($null -eq $driftRoot -or -not (Test-Path -Path $driftRoot)) {
         return $null
     }
 
-    # Get root directory for drift tests
-    # This assumes the drift tests are located in a folder named "drift" at the root of the current location
-    $driftRoot = Join-Path -Path $(Get-Location) -ChildPath "drift"
-    # Ensure the drift root directory exists
-    if (-not (Test-Path -Path $driftRoot)) {
+    # Ensure the Compare-MtJsonObject function is available
+    if (-not (Get-Command -Name Compare-MtJsonObject -ErrorAction SilentlyContinue)) {
+        Write-Warning "Compare-MtJsonObject function missing, not the right version of Maester?"
         return $null
     }
 


powershell/public/maester/drift/Compare-MtJsonObject.ps1
```diff
@@ -0,0 +1,164 @@
...
+            PropertyName = $Path
+            ExpectedValue = $Baseline
+            ActualValue = $null
+            Description = "Current is null at path: $Path"
+            Reason = "NullCurrent"
+        }
+        $differences += $issue
+        return $differences
+    }
+    if ($Baseline -is [System.Collections.IDictionary] -or $Baseline -is [PSCustomObject]) {
```
moorereason: Can't help but mention a pet peeve: reverse the logic of this if-statement and return early (e.g., `if (NOT this -and NOT that){ return }`).  This will remove a level of indentation from the rest of the function.

svrooij: Would you be able to elaborate?

Would that work with the recursive checking?
I mean if it is both an array, it would have to loop through the array to test things.
Not sure if PowerShell would count two arrays as " being equal".

And this is just a first try, I'm hoping that people would test this new feature and provide feedback. We can always make it better 🤩

moorereason: The functionality shouldn't change.  I'm only trying to remove cognitive load from unnecessary indentation.

```powershell
# Negate the original logic of this if-statement at line 84 and return early
if ($Baseline -isnot [System.Collections.IDictionary] -and $Baseline -isnot [PSCustomObject]) {
    return $differences
}
# Now we need less indentation for the remaining code
$properties = if ($Baseline -is [System.Collections.IDictionary]) { $Baseline.Keys } else { $Baseline.PSObject.Properties.Name }
# and so on...
```
Does that make more sense?

svrooij: I changed the code a bit, it seems to still work. Is this better?

moorereason: Yes, that's great. :+1:

merill:
I like the idea and I think it can be a powerful toolkit in the Maester arsenal.

Here's how I feel we can evolve this. 

Instead of a single test like MT.1060 how about making all the functionality available as a single cmdlet.

I'm thinking we move most of the logic in the 1060.Tests.ps1 file  into the cmdlet (let's call it Compare-MtObject)

Now, whenever someone wants to do a drift comparison in a custom test, they can use this cmdlet. 

**MVP 1** 
`Compare-MtObject -Path ./path1/myfile.json -ComparePath ./path2/myfile.json`

This will compare the two json files and fail if they are different

**MVP 2**
`Compare-MtObject -Path ./path1/myfile.json -CompareGitHistory`

`-CompareGitHistory` This will assume the file is in a git repo and check with the previous version of the file for diff

**MVP 3**
`Compare-MtObject -Path ./path1/ -ComparePath ./path2/` 

This compares an entire folder, look for json files and compare them

Since it's and Mt cmdlet we can make it call Add-MtTestResultDetail so it automatically provides all the info and have additional options like creating multiple test results for each file.

A good example we can showcase is a test that uses [EntraExporter](https://aka.ms/EntraExporter) then performs a CompareMt-Object to check for diff with a previous version. 

Another test could be using M365Dsc (and somehow extracting that data), then doing a diff to compare.

svrooij:
@merill that are some great ideas, but I wanted to avoid that people would need to create custom tests to use this functionality. They would all look the same, I guess?

Which is also why the Compare-MtJsonObject is exposed as a public function ✌️you can still make your own tests with this

SamErde:
How might `Compare-MtTestResult` fit into this conversation? It could use some refinement and a custom format, but I think it can at least cover the functionality I was hoping for in previous conversations. That said, the need to monitor configuration drift is still crucial!

For the sake of discussion, I wonder:

- does this align with the core needs that Maester is trying to fulfill?
- do we want to expand Maester into a configuration validation, hardening, and tracking suite?
- do other tools already do this (or almost do this?)

I'm not asking these questions to suggest an answer, but to genuinely find out what, if any, defined scope Maester might exist within. 👍✌️

powershell/Maester.psd1
```diff
@@ -78,7 +78,7 @@ FormatsToProcess = @('Maester.Format.ps1xml')
 # Functions to export from this module, for best performance, do not use wildcards and do not delete the entry, use an empty array if there are no functions to export.
 FunctionsToExport = 'Add-MtTestResultDetail',
     'Clear-MtDnsCache', 'Clear-MtExoCache', 'Clear-MtGraphCache',
-    'Compare-MtTestResult',
+    'Compare-MtJsonObject', 'Compare-MtTestResult',
```
SamErde, line 81: As an aside, much 🤩 for adding `Compare-MtJsonAobject` precisely here! 😂

powershell/public/maester/drift/Compare-MtJsonObject.ps1
SamErde, line 1: This could definitely be a useful script! Do you have a clear distinction for when this might be used and when `Compare-MtTestResult` might be used?

powershell/Maester.psd1
svrooij: Alphabetically is the best way ever

merill:
That's an excellent suggestion @soulemike. I'm a fan of the idea of having a lot of plug-in provider options.

Should the part of calling the provider to download to the local folder and then write back to the provider all be bundled into Invoke-Maester? Or should they be seperate steps/cmdlets.

svrooij:
This pr is just auto discovering files in a specific folder. Allowing every user to do drift tests, while managing the storage of the files in a way they prefer. Generating at run time, downloading from an url or getting them from git.

There are no git commands here just yet. And I don't want to wait until we design a "plugin-system" for storage of data between runs and using that for drift tests. I think we should create a new issue (discussion??) for "how do we manage data between runs, to show changed results, and drift checking".

soulemike:
> This pr is just auto discovering files in a specific folder. Allowing every user to do drift tests, while managing the storage of the files in a way they prefer. Generating at run time, downloading from an url or getting them from git.
> 
> There are no git commands here just yet. And I don't want to wait until we design a "plugin-system" for storage of data between runs and using that for drift tests. I think we should create a new issue (discussion??) for "how do we manage data between runs, to show changed results, and drift checking".

Yeah, and this was the point I was attempting to get to. The module itself is stateless. If there is no method within the module to manage state then every user will implement it in whatever way they want. We can either provide a recommended way to manage state within the module or just let them handle it as they want outside of the module.

merill:
@svrooij Some additional thoughts

**Drift folder location**
The `drift` folder should be a sub-folder of the `Custom` folder. When users run `Update-MaesterTests` all other folders are removed.

**Folder comparison**
The current implementation uses a baseline.json and a current.json. This feels very specific and would require some work from users to rename their files to match what we have here.

Since this is going to be a generic test, I feel a more flexible option is to compare with a folder structure:

```
- tests
	- custom
		- drift
			- baseline
			- current
```

**Docs**
Since this is going to be out of the box and requires customizing, we should add a section for Drift tests (maybe below the Custom section). Since Maester's focus has always been to make it easy for beginners we should provide a useful example they can follow to get a working drift comparison.

I feel something like aka.ms/EntraExporter with Domains being exported as a `baseline` and being compared with `current` for each run. This would be an useful example to fail when someone adds/removes or makes changes to a tenant's list of Domains. (thoughts @SamErde)

SamErde:
> **Drift folder location** The `drift` folder should be a sub-folder of the `Custom` folder. When users run `Update-MaesterTests` all other folders are removed.

Great call-out from a data preservation standpoint! Would there be value in keeping data separate from code? Is there a principle that should be followed here? _If so_, it would introduce a small new challenge, but not a bad one: a default location for test results. (Sorry, I'll get to your actual question, too!) 😅

Setting a standard location for Maester data (overridable with the **OutputFolder** parameter) could provide some helpful structure for users/admins. Also, with a default location for configuration and tests, the Maester framework itself could more reliably target known locations with new and improved functionality (eg: updating tests). The XDG base directory specification (and our best approximation on Windows) could be a good, cross-platform starting point for both.

**Content Types** | **OS** | **Folder** |
| :----------- | :----- | :----------|
| Module configuration
| | Linux | ~/.config/Maester
| | Windows | AppData/Maester/config
| Test results, drift data
| | Linux, macOS |
| | | ~/.local/share/Maester/test-results
| | | ~/.local/share/Maester/drift
| | Windows |
| | | LocalAppData/Maester/test-results
| | | LocalAppData/Maester/drift

I think this would be fairly simple to setup and would then provide a consistent framework for things like Entra Exporter that add more data.

> I feel something like aka.ms/EntraExporter with Domains being exported as a baseline and being compared with current for each run. This would be an useful example to fail when someone adds/removes or makes changes to a tenant's list of Domains. (thoughts @SamErde)

I'm unsure about the idea of throwing a _fail_ when a domain is added or removed in a tenant. Would that be to flag security needs or just to indicate that a change/drift has been detected? My personal preference would be to not use the pass/fail nomenclature for drift detection, but to instead include these in a separate section or a separate report with tags such as '➕ Added', '〽️ Changed', or '➖ Removed' for reported drift.

However, I _do_ like the idea of calling out configuration changes that could represent a significant change to the posture of the tenant, as these could affect (or require changes to) things like conditional access policies. In that vein, drift reporting could use the 'ℹ️ Info' label and 'Configuration Drift' category, as the 'Info' label isn't used much yet.

Now, this might be worth spinning off into a separate discussion, but Entra Exporter provides some really interesting possibilities:

- We could use Entra Exporter to facilitate additional configuration data for drift monitoring, _or_ the drift monitoring feature could be added to Extra Exporter.
- Data from Entra Exporter could be used to enhance or clarify drift monitoring insights [within Maester].
- We could add a feature for Maester to perform "offline assessments" against a data set provided by Entra Exporter (*\*nods in acknowledgement to @soulemike\**). 🙂

Thoughts on any of the above?

svrooij:
@SamErde I'm really not a fan of the defaults you're suggesting, in my vision Maester is run in some automation 90% of the time. I have no stats, but I cannot imagine that most Maester runs are done from an interactive terminal. What about we do a parameter `-ConfigPath`, with some sensible defaults, maybe changing per platform. This is what you would then use in for instance the GitHub action to set to a github action specific path.

Maybe appdata is good for storage when running on someone his computer. But using that on other platforms is problamatic.

Regarding this PR, the drift tests, is there a way to make a config value that you pass to `Invoke-Maester` like `-DriftPath` to the pester test? If there a way to use config variables from the pester config?

SamErde:
> @SamErde I'm really not a fan of the defaults you're suggesting, in my vision Maester is run in some automation 90% of the time. I have no stats, but I cannot imagine that most Maester runs are done from an interactive terminal. What about we do a parameter `-ConfigPath`, with some sensible defaults, maybe changing per platform. This is what you would then use in for instance the GitHub action to set to a github action specific path.
> 
> Maybe appdata is good for storage when running on someone his computer. But using that on other platforms is problamatic.

Now I'm curious. I assumed that at least a simple majority of Maester _users_ are actually running it interactively even though there is probably a large majority of Maester _runs_ that are actually done as a part of an automation job.

I do have a pretty comprehensive script that checks if running in an interactive session, non-interactive, or even within a CI/CD pipeline.
https://gist.github.com/SamErde/4ae30a56f8a805507b017212b811f64f

merill:
I know we called it tests, but in my mind that folder is the root for the current test run.

Users could have different Maester folders for each tenant they work with. I want to avoid adding complexity, unless the user themselves choose to use different folder.

So for defaults I prefer using the /Custom folder and maybe reserving /drift under that as the default for drift tests.

svrooij:
@merill and @SamErde what is the best way to make drift tests configurable?
I was thinking about an environment variable, something like `MEASTER_FOLDER_DRIFT`?

This allows for total flexibility, and allows you to run these tests on demand. You have to do something for them anyway. You can also set environment variables from running scripts right? That in Invoke-Measter it would have a `-DriftFolder` parameter that would then set this environment variable.

Custom PesterConfiguration value is also fine by me, but cannot find any docs about that.

powershell/public/maester/drift/Compare-MtJsonObject.ps1
svrooij: It has documentation included? Any improvements or should this be placed somewhere else?

powershell/public/Invoke-Maester.ps1
```diff
@@ -365,6 +369,21 @@ function Invoke-Maester {
...
+    if ($DriftRoot) {
+        $DriftRoot = (Resolve-Path -Path $DriftRoot -ErrorAction SilentlyContinue).Path
+        if (-not (Test-Path -Path $DriftRoot)) {
+            Write-Warning "❌ The specified drift root directory '$DriftRoot' does not exist." -ForegroundColor Red
+
+        } else {
+            Set-Item -Path Env:\MEASTER_FOLDER_DRIFT -Value $DriftRoot
+            Write-Verbose "🧪 Drift root directory set to: $DriftRoot"
+        }
+    } else {
```
SamErde, line 382: Do you want to remove the empty `else` block and extra line break above, or are they placeholders for future work?

merill (approved):
LGTM with DriftRoot as a parameter

powershell/public/Invoke-Maester.ps1
merill: Left in as a comment for now. Can be cleaned up.

PR #996, opened 2025-06-26
Fix validation that forces guid for mailUserId in SendMtMail when up also works.

maikvandergaag:
Added validation to the MailUserId parameter as it is required to be a guid

### DIFF
--- a/powershell/public/Invoke-Maester.ps1
+++ b/powershell/public/Invoke-Maester.ps1
@@ -145,15 +145,8 @@ function Invoke-Maester {
         # Uri to the detailed test results page.
         [string] $MailTestResultsUri,
 
-        # The user id of the sender of the mail. Defaults to the current user.
+        # The user id or user principal name of the sender of the mail. Defaults to the current user.
         # This is required when using application permissions.
-        [ValidateScript({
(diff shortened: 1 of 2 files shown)


merill:
Thanks for the PR @maikvandergaag 

This is actually a good catch. The UserId parameter can be a upn too.

I just tested this succesfully.

`Invoke-Maester -Tag 'MT.1058' -MailUserId merill@elapora.com -MailRecipient merill@elapora.com`

We accidentally added a guid check to send-mtmail so what I'm going to do is remove the validation altogether and call out that UPN can work too.

merill:
I've pushed the update. Can you please test and review if it works for you? Tx

maikvandergaag:
Thanks it works!

> Thanks for the PR @maikvandergaag
> 
> This is actually a good catch. The UserId parameter can be a upn too.
> 
> I just tested this succesfully.
> 
> `Invoke-Maester -Tag 'MT.1058' -MailUserId merill@elapora.com -MailRecipient merill@elapora.com`
> 
> We accidentally added a guid check to send-mtmail so what I'm going to do is remove the validation altogether and call out that UPN can work too.

Yes we noticed that and that was why our automation failed. Cool that it is now working how it should.

PR #1004, opened 2025-06-30
Adding new test MT.1063 - All App Registration owners should have MFA set up

marcel-ngn:
This PR adds a new test: MT.1063 - All App Registration owners should have MFA set up
It checks all owners of app registration and the MFA status of the owner. The test will fail if an owner does not have MFA set up.


merill (changes requested):
Please see comments above

marcel-ngn:
Hi @merill 
Following changes were made to optimize the runtime:
1. Using a single Graph call to get the app registrations and their owner(s). However I was not able to filter for only apps with owners on server side. In my latest version im still filtering for apps with owners on client side. Do you know a trick here? :) 

2. MFA check: Instead of checking every individual users MFA status via Graph, Im now using the userRegistrationDetails API to get the MFA status of all users and saving the result to a table for more efficiency (using hashSet to ensure every owner is only getting checked once for MFA).

Bildschirmfoto 2025-07-08 um 08 36 24

Bildschirmfoto 2025-07-08 um 08 42 25


Additions: If an owner could not be checked (e.g. disabled user or service principal) it will show the details of that object.

Bildschirmfoto 2025-07-08 um 08 38 26

marcel-ngn:
- fixed a typo in the function name

merill:
Made a few updates to simplify the summary message. It was getting a little confusing with all the numbers.

Also updated the urls so they will work with gov, China and other clouds.

PR #1005, opened 2025-06-30
fix&improve MtCisaAuditLogPremium Test

kentbjoh:
This PR fix and improves the CisaAuditLogPremium Test

- Fetch all actual mailboxes
- Pass correct results to $testResult (if 0 results, its a pass)
- Show some quick info on how many has passed or not
- Limit results view to 100


merill:
Thanks @kentbjoh for the PR. 

However I looked up the CISA docs and this test appears to have been deprecated now thanks to changes by Microsoft. 

Please see https://github.com/cisagov/ScubaGear/blob/7cefa12639b4bc36990f8f2849b57ad2fdafec4c/PowerShell/ScubaGear/baselines/removedpolicies.md?plain=1#L56

@soulemike  can you review this as well?

I've updated this PR to one that deprecates the test (to be removed at a later stage).

LMK if you think this test is still valuable outside of CISA recommendation?

PR #1014, opened 2025-07-08
Fix tests CIS.M365.3.1.1 and CIS.M365.2.1.12

ic3lotus:
**Test: CIS.M365.3.1.1**
Remove selection of property to not lose any other properties used in the foreach loop later (e.g. id).

**Test: CIS.M365.2.1.12**
Switch from the name of the variable to the automatic variable which contains the pipeline object and add the arraylist name via dot-sourcing in the conditional expression to properly read the count. Without adding the name of the arraylist, the count would always show 1 because the IPAllowList does count as one item as well)


merill (approved):
Good catch. Thanks for the PR @ic3lotus

Issue #1020, opened 2025-07-11
🪲 The action parameter 'mail_userid' only works with a GUID now when it should also accept a UPN


wmacdonald10:
Thanks for reporting the bug. Please ensure you've gone through the following checklist before opening an issue:

- Make sure you can reproduce this issue using the latest released version of `Maester`.
- Please search the existing issues to see if there has been a similar issue filed.

## Describe the bug
The Send-MTMail.ps1 PowerShell script was updated to only accept a GUID for the UserId parameter, the UserId parameter should accept a UPN & GUID as in its current state it is a breaking change for anyone who used a UPN value for mail_userid

## To Reproduce

1. Configure the action and set  mail_userid to the UPN of the user
2. See error ` Write-Error: /home/runner/work/_temp/592f78ed-72a5-4a7a-8c62-7ca45e50d577.ps1:2
Line |
   2 |  /home/runner/work/_actions/maester365/maester-action/v1.0.1\script\Ru …
     |  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | Failed to run Maester tests. Please check the parameters. Cannot
     | validate argument on parameter 'UserId'. Invalid UserId format. It
     | should be a valid GUID. at         $results = Invoke-Maester
     | @MaesterParameters  in
     | /home/runner/work/_actions/maester365/maester-action/v1.0.1/script/Run-MaesterAction.ps1
Error: Process completed with exit code 1.`

## Expected behavior
Microsoft Graph supports the id or UPN for the [SendMail](https://learn.microsoft.com/en-us/graph/api/user-sendmail?view=graph-rest-1.0&tabs=http) Endpoint 

I will attach a fix for Send-MtMail.ps1

## Debug Output

⚠ ATTENTION: Be sure to remove any sensitive information that may be in the logs.

<details>
<summary>Debug Output</summary>

</details>

wmacdonald10:
If you're impacted by this bug, you can quickly just lookup the Object ID for the user account and use that instead for the time being.

merill:
Thanks for the PR. Should be fixed now

PR #1021, opened 2025-07-11
UserId parameter should accept a UPN or Object ID

wmacdonald10:
Fix for Issue #1020 

Updated parameter UserId validate script to accept a UPN


merill (approved):
Thanks for the catch!

PR #1027, opened 2025-07-14
refactor: improve module loading with better error handling

SoufiyanAk:
- Add graceful handling for missing script files
- Replace hard errors with warnings to prevent module load failures
- Improve error messages with more context and file paths
- Add safe module manifest loading with fallback
- Track and report import errors for better debugging

### DIFF
--- a/powershell/Maester.psm1
+++ b/powershell/Maester.psm1
@@ -25,15 +25,34 @@ $__MtSession = @{
 New-Variable -Name __MtSession -Value $__MtSession -Scope Script -Force
 
 # Import private and public scripts and expose the public ones
-$privateScripts = @(Get-ChildItem -Path "$PSScriptRoot\internal" -Recurse -Filter "*.ps1")
-$publicScripts = @(Get-ChildItem -Path "$PSScriptRoot\public" -Recurse -Filter "*.ps1")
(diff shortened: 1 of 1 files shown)


powershell/Maester.psm1
```diff
@@ -25,15 +25,34 @@ $__MtSession = @{
...
 
-$ModuleInfo = Import-PowerShellDataFile -Path "$PsScriptRoot/Maester.psd1"
\ No newline at end of file
+# Report import errors if any occurred
+if ($importErrors.Count -gt 0) {
+	Write-Warning "Module loaded with $($importErrors.Count) import error(s). Some functionality may be unavailable."
+}
+
+# Safely import module manifest
+try {
```
SamErde, line 53: This `try/catch` block _looks_ redundant to me because it returns the same warnings and $ModuleInfo as when the statement is not wrapped.

merill: It also might hide some of the errors since we are only showing the message. but I'm okay with this for now. I'd like to integrate something like PSFramework so we can log these detailed errors and be able to to export it with PSFramework cmd. Idea for a future update.

merill (approved):
LGTM thanks @SoufiyanAk

PR #1029, opened 2025-07-14
refactor: enhance progress display with better error handling

SoufiyanAk:
- Add safe host width detection with fallback to prevent errors
- Improve macOS compatibility with more targeted workaround using $IsMacOS
- Add graceful fallback to Write-Host when Write-Progress fails
- Optimize string handling for status messages with proper trimming
- Better error logging with Write-Debug instead of Write-Verbose

### DIFF
--- a/powershell/internal/Write-MtProgress.ps1
+++ b/powershell/internal/Write-MtProgress.ps1
@@ -31,32 +31,52 @@ function Write-MtProgress {
         $Activity = "🔥 $Activity"
 
         if ($Status) {
-            $statusString = Out-String -InputObject $Status
+            $statusString = if ($Status -is [string]) { $Status } else { Out-String -InputObject $Status }
+
+            # Safely get host width with fallback
+            $hostWidth = 80 # Default fallback
+            try {
+                if ($Host.UI.RawUI.WindowSize) {
+                    $hostWidth = $Host.UI.RawUI.WindowSize.Width
+                }
+            } catch {
+                Write-Debug "Unable to get host width, using default: $_"
+            }
+
             # Reduce the length of the status string to fit within host
-            $hostWidth = $Host.UI.RawUI.WindowSize.Width
             $buffer = 20
-            $totalWidth = $Activity.Length + $statusString.Length + $buffer # 10 for buffer
+            $totalWidth = $Activity.Length + $statusString.Length + $buffer
             if ($totalWidth -gt $hostWidth) {
                 $length = $hostWidth - $Activity.Length - $buffer
-                if ($length -lt $statusString.Length -and $length -gt 0) {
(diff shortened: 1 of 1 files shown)


powershell/internal/Write-MtProgress.ps1
```diff
@@ -31,32 +31,52 @@ function Write-MtProgress {
         $Activity = "🔥 $Activity"
 
         if ($Status) {
-            $statusString = Out-String -InputObject $Status
+            $statusString = if ($Status -is [string]) { $Status } else { Out-String -InputObject $Status }
```
SamErde, line 35: Nice trick. Have you ever seen $Status come back as a non-string or is this just a good practice to follow?

powershell/internal/Write-MtProgress.ps1
```diff
@@ -31,32 +31,52 @@ function Write-MtProgress {
...
-            $hostWidth = $Host.UI.RawUI.WindowSize.Width
             $buffer = 20
-            $totalWidth = $Activity.Length + $statusString.Length + $buffer # 10 for buffer
+            $totalWidth = $Activity.Length + $statusString.Length + $buffer
             if ($totalWidth -gt $hostWidth) {
                 $length = $hostWidth - $Activity.Length - $buffer
-                if ($length -lt $statusString.Length -and $length -gt 0) {
-                    $statusString = $statusString.Substring(0, $length) + "..."
+                if ($length -gt 0 -and $length -lt $statusString.Length) {
+                    $statusString = $statusString.Substring(0, $length).TrimEnd() + "..."
```
SamErde, line 53: Curious because I like to learn and also try to keep the code simple: could this just be `$statusString = $statusString.TrimEnd() + "..."` or is the substring required for some cases?

SamErde (approved):
Nice updates.

powershell/internal/Write-MtProgress.ps1
SoufiyanAk: In most cases, $Status is a string, but I’ve seen situations (especially when passing objects or arrays by mistake) where it could be something else...
Adding the check is mostly just a defensive habit to avoid weird output or errors if someone ever passes a non-string value. It’s probably not strictly necessary here, but I like to play it safe

powershell/internal/Write-MtProgress.ps1
SoufiyanAk: The Substring is there to make sure the status string doesn’t go past the available width in the terminal, so it trims the string to fit before adding the ellipsis. If we just used TrimEnd() + "...", the string could still be too long and might wrap or get cut off awkwardly in the console :)

SamErde (approved):

PR #1030, opened 2025-07-14
refactor: improve Get-MtGroupMember with better validation

SoufiyanAk:
- Add ValidateNotNullOrEmpty attribute for GroupId parameter
- Implement empty GUID validation to prevent invalid requests
- Enhance error messages with more context and exception details
- Improve recursive group handling with better error recovery
- Add comprehensive null checks and validation throughout function
- Better verbose logging with group display name for debugging

### DIFF
--- a/powershell/public/Get-MtGroupMember.ps1
+++ b/powershell/public/Get-MtGroupMember.ps1
@@ -15,34 +15,63 @@ function Get-MtGroupMember {
   [CmdletBinding()]
   param(
     [Parameter(Position = 0, mandatory = $true)]
+    [ValidateNotNullOrEmpty()]
     # ID for the Entra group to return members for.
     [guid]$GroupId,
     # Include indirect members through nested groups.
     [switch]$Recursive
   )
 
-  try {
-    Invoke-MtGraphRequest -RelativeUri "groups/$GroupId/" -ApiVersion v1.0 | Out-Null
-  } catch {
-    Write-Error "Error obtaining group ($GroupId) from Microsoft Graph. Confirm the group exists in your tenant."
+  # Validate GroupId is not empty GUID
+  if ($GroupId -eq [guid]::Empty) {
+    Write-Error "GroupId cannot be an empty GUID."
     return $null
   }
 
-  Write-Verbose -Message "Getting group members."
+  try {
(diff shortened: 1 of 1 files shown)


powershell/public/Get-MtGroupMember.ps1
```diff
@@ -15,34 +15,63 @@ function Get-MtGroupMember {
   [CmdletBinding()]
   param(
     [Parameter(Position = 0, mandatory = $true)]
+    [ValidateNotNullOrEmpty()]
     # ID for the Entra group to return members for.
```
SamErde: IMHO, I find it to be more helpful if the comment is placed above the parameter definition that it is describing vs directly above the parameter type and name, but that may just be a matter of personal preference. 🙂

powershell/public/Get-MtGroupMember.ps1
```diff
@@ -15,34 +15,63 @@ function Get-MtGroupMember {
...
   }
 
-  Write-Verbose -Message "Getting group members."
+  Write-Verbose -Message "Getting group members for group: $($group.displayName) ($GroupId)"
 
-  $members = @()
-  $members += Invoke-MtGraphRequest -RelativeUri "groups/$GroupId/members" -ApiVersion v1.0
+  try {
+    $members = @()
+    $groupMembers = Invoke-MtGraphRequest -RelativeUri "groups/$GroupId/members" -ApiVersion v1.0
```
SamErde: Curous: could this just be `$members = Invoke-MtGraphRequest -RelativeUri "groups/$GroupId/members" -ApiVersion v1.0` and skip the `if ($groupmembers)` statement below?

powershell/public/Get-MtGroupMember.ps1
SoufiyanAk: I agree, placing the comment above the parameter definition does make it clearer.

powershell/public/Get-MtGroupMember.ps1
SoufiyanAk: My original thought was to handle the case where the request might return nothing, but you’re right—assigning $members directly is simpler and works just as well

SamErde (approved):

PR #1038, opened 2025-07-17
Refactor Connect-MgGraph to support optional GraphApplicationId

kevinstone-at-res:
The change allows users to create a custom app for Graph connection and avoid the OOB Graph CLI app, which can become over-privileged. 

1. Added [string]$GraphClientId param to Connect-Maester
2. Made graph connection logic more DRY while incorporating $GraphClientId

### DIFF
--- a/powershell/public/Connect-Maester.ps1
+++ b/powershell/public/Connect-Maester.ps1
@@ -63,9 +63,9 @@
    Connects to China environments for Microsoft Graph, Azure, and Exchange Online.
 
 .EXAMPLE
-Connect-Maester -TenantId "4e285730-281f-42a2-bdfd-e766394e85d0" -GraphClientId "f45ec3ad-32f0-4c06-8b69-47682afe0216"
+   Connect-Maester -GraphClientId 'f45ec3ad-32f0-4c06-8b69-47682afe0216'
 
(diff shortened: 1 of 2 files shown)


merill (changes requested):
@kevinstone-at-res This is a good addition. Can I make some suggestions?

- [ ] Rename to ClientId so we are consistent with `Connect-Graph`
- [ ] Include an example in the .ps1 file that shows a custom ClientId being used
- [ ] In the param reference point to the Microsoft doc on how to create a custom app for this (https://learn.microsoft.com/en-us/powershell/microsoftgraph/authentication-commands?view=graph-powershell-1.0#use-delegated-access-with-a-custom-application-for-microsoft-graph-powershell)
- [ ] Update website/docs/connect-maester/readme.md and add a section that describes how to use a custom client (point to the Microsoft doc above on how to create a custom app for Graph PS).
- [ ] We don't need to include a warning if no TenantId is specified and common is used. That's expected behaviour.

kevinstone-at-res:
@merill Thank you for the review — I’ve implemented all the changes.

I also wanted to note that since the documentation states Connect-Maester is completely optional, I aimed to keep this change lightweight and compatible with that guidance. If the team decides not to include it, that’s totally understandable. We can call Connect-MgGraph directly with whatever options we need.

PR #1040, opened 2025-07-20
Exchange SPF - Adjust Error Handling

weyCC81:
I mostly see this failure when the SPF can not even be caught/requested


merill (approved):
Thanks @weyCC81

PR #1051, opened 2025-07-23
Add new test: MT.1068: Restrict non-admin users from creating tenants.

marcel-ngn:
Adding a new test that checks if non-admin useres are restricted from creating new tenants. Check is done via policies/authorizationPolicy?$select=defaultUserRolePermissions

image

image


merill:
Good stuff. Thanks a lot @marcel-ngn

Issue #1069, opened 2025-08-01
🪲 MS.EXO.2.2 - Stack overflow if SPF records have recursive loop


mattcave:
## Describe the bug

When running the MS.EXO.2.2 (SPF) test against a domain that has a recursive loop in the SPF records (i.e. it `include`s itself), the test fails to identify the misconfiguration and goes into an infinite loop, leading to a stack overflow. This results in the execution aborting and other tests not being run. 

## To Reproduce


1. `Invoke-Maester -Tag MS.EXO.2.2`
2. Wait a while
3. Stack overflow error


## Expected behavior

Maester should catch this misconfiguration and report it as a problem, but not "explode"


## Debug Output
<!-- 🔬 Run the problematic command with `-Debug` and paste the resulting debug stream below. -->
⚠ ATTENTION: Be sure to remove any sensitive information that may be in the logs.

<details>
<summary>Debug Output</summary>

```
Discovery found 264 tests in 21.74s.                                                                                    
Filter 'Tag' set to ('MS.EXO.2.2').                                                                                     
Filter 'ExcludeTag' set to ('CAWhatIf').                                                                                
Filters selected 1 tests to run.                                                                                        
Running tests.                                                                                                          
                                                                                                                        
Running tests from '/Users/matt/Code/Augmentt/maester/tests/cisa/exchange/Test-MtCisaSpfDirective.Tests.ps1'
Describing CISA SCuBA                                                                                                   
  [-] MS.EXO.02.2: An SPF policy SHALL be published for each domain, designating only these addresses as approved senders. 36.81s (36.8s|14ms)
   ScriptCallDepthException: The script failed due to call depth overflow.                                              
   RuntimeException: The script failed due to call depth overflow.
   at Resolve-SPFRecord<Process>, /Users/matt/Code/Augmentt/maester/powershell/public/cisa/exchange/Resolve-SPFRecord.ps1:71
   at Resolve-SPFRecord<Process>, /Users/matt/Code/Augmentt/maester/powershell/public/cisa/exchange/Resolve-SPFRecord.ps1:147
   at Resolve-SPFRecord<Process>, /Users/matt/Code/Augmentt/maester/powershell/public/cisa/exchange/Resolve-SPFRecord.ps1:147
   at Resolve-SPFRecord<Process>, /Users/matt/Code/Augmentt/maester/powershell/public/cisa/exchange/Resolve-SPFRecord.ps1:147
[...]

PR #1070, opened 2025-08-01
Handle SPF records that include themselves; prevent an infinite loop

mattcave:
When running the MS.EXO.2.2 (SPF) test against a domain that has a recursive loop in the SPF records (i.e. it includes itself) , the test fails to identify the misconfiguration and goes into an infinite loop, ultimately resulting in a stack overflow. This results in the execution aborting and other tests not being run.

This PR adds a simple check in the Resolve-SPFRecord function to prevent this

https://github.com/maester365/maester/issues/1069


merill:
Awesome. Thanks @mattcave and @soulemike

Issue #1064, opened 2025-07-25
🪲 MT.1053 Ensure Itnune Device cleanup fail


jkerai1:
I used the previous ALl Devices clean up rule and this was imported into the new Rule engine but still showing as fail:

<img width="1985" height="708" alt="Image" src="https://github.com/user-attachments/assets/9c87aa2c-271d-4f2d-9dfd-7e274444ecea" />

l-gosling:
@jkerai1 The Bug is solved. If you want you can use the preview version of maester to test it.

merill:
Thanks for the fix and follow up @l-gosling

Issue #1058, opened 2025-07-24
🪲 MT.1053 reports "Test failed" although rule is in place


GeldHades27355:
## To Reproduce

1. install maester 1.3.0
2. autoinstall pester 5.5.0
3. connect-azaccount -tenant -scope
4. invoke-maester

## Expected behavior
Rule exists, so result should be PASS

## Environment Data
PSVersion                      7.5.2
PSEdition                      Core
GitCommitId                    7.5.2
OS                             Microsoft Windows 10.0.26100, Arm64
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
Tenant license M365 BP

GeldHades27355:
Sorry, forgot the error thrown: Test-MtManagedDeviceCleanupSettings: C:\temp\maester\Maester\Intune\Test-MtIntunePlatform.Tests.ps1:3
Line |
   3 |          $result = Test-MtManagedDeviceCleanupSettings
     |                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | Response status code does not indicate success: Forbidden (Forbidden).

merill:
Thanks folks, this has been merged in. Available in the latest preview build.

PR #1080, opened 2025-08-06
updated to new api url and to new interface in intune

l-gosling:
Updated to new api url and to new interface in intune.
added output as table if multiple rules are defined.

Fix for https://github.com/maester365/maester/issues/1064 and https://github.com/maester365/maester/issues/1058

New failed output:
image

New passed output:
image


merill:
Nice! This looks much improved now. Thanks for the update.

I made a few tweaks to the docs!

PR #1082, opened 2025-08-08
Workaround for Microsoft bug in EOP\Get-AdminAuditLogConfig function

### LINKED ISSUE #1045: 🪲 Admin audit log check may incorrectly return $false for UnifiedAuditLogIngestionEnabled due to ExchangeOnlineManagement bug
Tests that rely on `Get-AdminAuditLogConfig` to ensure the unified audit log is enabled may incorrectly return a `$false` value due to a bug in the SCC (security and compliance center) portion of the **ExchangeOnlineManagement** module. Currently affected checks are:

- **CIS.M365.3.1.1**: (L1) Ensure Microsoft 365 audit log search is Enabled<br>cis\Test-MtCisAuditLogSearch.Tests.ps1

- **CISA.MS.EXO.17.1**: Microsoft Purview Audit (Standard) logging SHALL be enabled.<br>cisa\exchange\Test-MtCisaAuditLog.Tests.ps1

**Note:** This is a bug with the **ExchangeOnlineManagement** module and will not be fixed by the Maester team, but I want to track it here (1) for awareness as a known issue and (2) because we _can_ work around it if we want to.

**Note:**, the ExchangeOnlineManagement module does not ship with most of its cmdlets. Instead, it imports a temporary module with all of the Exchange Online management cmdlets after you connect to the online API endpoint using `Connect-ExchangeOnline`. It does this again, importing a 2nd temporary module when you use `Connect-IPPSSession`.

In short, a bug in the SCC sub-module (from the IP PS Session) breaks the `Get-AuditLogConfig` cmdlet. As you can see in the screen shot below:

Image

1. Connecting to Exchange Online correctly reports that I have the setting enabled.
2. Connecting after that to the SCC (IPPSSession) incorrectly reports that I do not have the setting enabled.

In `Connect-Maester`, we logically connect in this order, which results in a `$false` result whenever the `Get-AuditLogConfig` cmdlet is used to check **UnifiedAuditLogIngestionEnabled**.

However, if we connect to the SCC (IPPSSession) first, and _then_ connect to Exchane Online, the correct value is returned.

## Conclusion

I don't like having to depend so much on carefully nuanced order of connections, but do have a change drafted for the `Connect-Maester` function to work around this. It does work, but I need to wrap up a few final checks before submitting the workaround.

🙏 **Note:** Please help raise this issue up to the Exchange team so the actual issue in the **ExchangeOnlineManagement** module can be fixed.

SamErde:
This pull request introduces a targeted fix to the `Connect-Maester.ps1` script, addressing an issue where the `Get-AdminAuditLogConfig` cmdlet becomes broken due to a temporary PSSession module imported by `Connect-IPPSSession`. The new logic ensures the correct version of the cmdlet is restored for Exchange Online (EXO) sessions.

Fixes #1045!

**Admin Audit Log Fix:**

* Added a script block that detects when an EOP session is connected, removes the broken `Get-AdminAuditLogConfig` function, and re-imports the correct EXO module to restore functionality.

### DIFF
--- a/powershell/public/Connect-Maester.ps1
+++ b/powershell/public/Connect-Maester.ps1
@@ -213,11 +213,12 @@
                block removes the broken function and re-imports the temporary PSSession module for EXO, which restores
                the working Get-AdminAuditLogConfig function.
             #>
-            if (Get-ConnectionInformation | Where-Object { $_.IsEopSession -eq $true -and $_.State -eq 'Connected'}) {
+            $ExchangeConnectionInformation = Get-ConnectionInformation
+            if ($ExchangeConnectionInformation | Where-Object { $_.IsEopSession -eq $true -and $_.State -eq 'Connected' }) {
                try {
                   # Remove the broken cmdlet and re-import the working EXO one.
                   Remove-Item -Path 'Function:\Get-AdminAuditLogConfig' -Force -ErrorAction SilentlyContinue
-                  Get-ConnectionInformation | Where-Object { $_.IsEopSession -ne $true -and $_.State -eq 'Connected' } |
+                  $ExchangeConnectionInformation | Where-Object { $_.IsEopSession -ne $true -and $_.State -eq 'Connected' } |
                      Select-Object -ExpandProperty ModuleName |
                         Import-Module -Function 'Get-AdminAuditLogConfig' > $null
                } catch {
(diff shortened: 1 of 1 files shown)


merill:
LGTM. Thanks Sam.

PR #1089, opened 2025-08-11
Hide empty results in detailed output of Test-MtConnection

SamErde:
Using `Test-MtConnection -Details` outputs a custom formatted list that shows the details of each service connection (Azure, Exo, Eop, Graph, Teams).

Specific service connections can be checked by using the **Service** parameter.

This change to the PowerShell format file for `Maester.Connections` dynamically hides null connections so it doesn't show services that you didn't ask it to test.


merill (approved):
Looks great. Thanks @SamErde

PR #1103, opened 2025-08-18
Fix for bug in older API

brianveldman:
In the older API the runbook association failed since couple weeks. In the newer API this is fixed.


brianveldman:
@merill Currently working on a new version of the deployment 🔥

PR #1104, opened 2025-08-18
Update Monitoring README

gabecorsini:
The BitBucket link showed GitLab in the alternate link text instead of BitBucket


merill (approved):
Good catch. Thanks @gabecorsini

PR #1108, opened 2025-08-20
MT.1073: Ensure soft and hard matching for on-premises synchronization objects is blocked

HenrikPiecha:
@merill Just worked on this test. After finalization, I noticed that the graph endpoint “directory/onPremisesSynchronization” does not support application permissions and I still had an active login with my ga. 
According to the documentation: [Update onPremisesDirectorySynchronization
](https://learn.microsoft.com/en-us/graph/api/onpremisesdirectorysynchronization-update?view=graph-rest-1.0&tabs=http)
In order not to lose the work, I wanted to add it anyway. Perhaps you have any idea on how to fix / workaround on this. Otherwise we'll need to drop this one.

Best regards
Henrik


merill:
This is still valuable and there will always be APIs that are delegate only. 

What I think we should do is check and only run in delegated and skip in app mode. Get-MgContext.AuthType should have the info.

merill (approved):
Good stuff. Thanks @HenrikPiecha I also made some tweaks to make this easier to follow for future devs!

tdcthosc:
Odd that the documentation says that Application permissions is not supported on GET, because I've just tested it and it works just fine... Perhaps improperly updated documentation?

merill:
Oh wait, if it does work then should we remove the check in this test?

PR #1110, opened 2025-08-20
GITLAB - WIF: Removed code to install ExchangeOnlineManagement preview. The stable release now support access token for IPPSSession

l-gosling:
Removed code to install ExchangeOnlineManagement preview. The stable release now support access token for IPPSSession.
At the time workload identity federation use was testest for maester this was only in the preview. Now it is in the stable release, so the install of the preview of ExchangeOnlineManagement if ipp connection should established is not necessary anymore.

Additional Changes:
Renamed "IPP" to "Purview".
Removed the "$AZURE_TENANT_NAME" and get it via graph instead.


merill (approved):
Perfect. Thanks!

PR #1112, opened 2025-08-21
Improve Invoke-MtGraphRequest

tdcthosc:
On a customer tenant with a large group, the Get-MtGroupMember went into a loop because the "SkipToken" in the Microsoft Graph "nextdata" was not unique after enough paging, meaning that Invoke-MtGraphRequest would find the "next page" in the cache, which 1) includes already collected members and 2) contain a previous "SkipToken". This would result in a loop, because that previous SkipToken URI was also in the cache etc.

@merill any idea why the "SkipToken" wouldn't be unique in a large call? The group has approximately 3700 users.

I've fixed it by not utilizing the cache when paging and also optimized performance by updated the initial call in the cache containing the full response. Each paging call is not saved in cache anymore – only that initial call.

I've also changed so that the results are only formatted once.

### DIFF
--- a/powershell/internal/Invoke-MtGraphRequestCache.ps1
+++ b/powershell/internal/Invoke-MtGraphRequestCache.ps1
@@ -41,7 +41,7 @@ function Invoke-MtGraphRequestCache {
             $results = Invoke-MgGraphRequest -Method $Method -Uri $Uri -Headers $Headers -OutputType $OutputType -Body $Body
         }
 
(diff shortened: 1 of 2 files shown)


merill (approved):
Thanks @tdcthosc this makes complete sense. 

It was very inefficient caching the paged results since they would never be used. This should also see a drop in memory used by Maester.

Great PR.

PR #1119, opened 2025-08-22
Html report cleanup

kij:
Missed a few spots going from `New-MtReport` -> `Export-MtHtmlReport` -> `Get-MtHtmlReport`.


merill:
Thanks a lot @kij Good updates.

PR #1127, opened 2025-08-25
Pr/htmlreportnotrun

kij:
### Description

1) This adds a NotRun card on the html summary

image

2) fixes a green->emerald in the donutChart

### Your checklist for this pull request

🚨Please review the [guidelines for contributing](https://maester.dev/docs/contributing) to this repository.

- [x] Read the guidelines for contributions

I've followed to documentation in` /report/README.md` - but I haven't updated the entire `const testResults` in  `App.jsx`, so that data no longer valid.  For trust reasons, I would like to think that it will be rebuild by whomever reviews this.  (Do we need a workflow for that now?)

#### PRs related to all code

- [x] Before you submit the PR, run the tests locally by running `/powershell/tests/pester.ps1`
- [x] After submitting, verify the tests are still passing on your PR in GitHub (the tests are run across all platforms).


report/src/components/MtTestSummary.jsx
```diff
@@ -64,6 +64,17 @@ export default function MtTestSummary(props) {
                 </Flex>
                 <ProgressBar value={getPercentage(props.SkippedCount)} color="gray" className="mt-3" showAnimation={true} />
             </Card>
+	    <Card>
+                <Flex alignItems="start">
+                   <Text>Not Run</Text>
```
SamErde: Nit pick: I like your use of Title Case for "Not Run," but for consistency with the casing of other card titles, it should be "Not run." (See "Not tested" and "Total tests.")

report/src/App.jsx
```diff
@@ -13,14 +13,15 @@ import MtBlocksArea from './components/MtBlocksArea';
 /*The sample data will be replaced by the Get-MtHtmlReport when it runs the generation.*/
```
SamErde, line 13: What are your thoughts on starting with "safe" defaults of 0 for failed, passed, skipped, not run, and even total tests--just in case the actual report does not populate properly for any unexpected reasons.

report/src/components/MtTestSummary.jsx
kij: Good catch! fixed 1afc6465605bf4912c5f732c63fc958a7edc3b81

powershell/assets/ReportTemplate.html
SamErde, line 1: I can't speak to the code change in ReportTemplate.html, but I can test it. What does this change do?

```
    let allSettled2 = function(promises) {
      return Promise.all(
        promises.map(
          (p2) => Promise.resolve(p2).then(
            (value) => ({ status: "fulfilled", value }),
            (reason) => ({ status: "rejected", reason })
          )
        )
      );
    };

```
and `promise = Promise.allSettled(`

Thanks!

report/src/App.jsx
kij: Right, 0  counts are better. Do you want to go all in an have default values for `TenantId/TenantName/Account` etc ?

SamErde: For now, let's keep changes in this PR related to your main goal: adding a 'Not run' card to the report so it's easier to review. Other changes can be submitted separately.

powershell/assets/ReportTemplate.html
kij: This is all one big artifact from the `npm run build  ; Copy-Item ./dist/index.html ../powershell/assets/ReportTemplate.html`

This is what I meant by trust. The build artifact - `ReportTemplate.html` should be created by a workflow or at least a maintainer. Too many ways to hide something malicious. 

As for why  it went from Promise.allSettled -> let allSettled2 = function(promises),  I don't really know.  I build Mac on with `nodejs-v24.6.0`
image

Again, you shouldn't really trust my `ReportTemplate.html`

SamErde: Do you want to just drop that commit, then?

kij: Well, I would prefer that you, or someone else overwrote it when the PR is accepted. I can drop it on my end,  but it's needed before the changes actually go through - So it should be bundled with the changes here :).

report/src/App.jsx
kij: give me a few hours, need to run an errand.

powershell/assets/ReportTemplate.html
kij: History seems to be that its updated with with the changes, as per: https://github.com/maester365/maester/pull/950  

It's still bad practice, and should be considered for update procedure :)

merill: Tx for raising this. I've updated the build process so the report is always built from source code when building the powershell module.

merill:
So I took a look into this PR. The report currently merges Skipped and Not Run into a single count as `Not Run`. This was intentional because I didn't see any value in showing these two different values. 

@kjj was there a specific reason you want to split out and show the skipped in the overview count?

On the whole, there is another state that we need to include in the report.

This is the count of tests that didn't complete because of an Error. E.g this can happen if the test code has a bug (we added try..catch blocks in all the tests). 

What I would really like to do is show this Error state in the test (instead of the Fail state that it does today) and then show an Error count at the top (if any tests Errored out). I feel this will have more value than splitting out Not Run and Skipped tests.

Thoughts?

merill:
Got it. I still want to understand the value you see in the distinction between NotRun and Skipped. While it makes sense in the Pester world, it's not something that I feel will be valuable to Maester end users that are looking at the report (especially the count at the top).

If there are errors it does get logged in the console.

In terms of tag hierarchy, I've never seen one that's been successful in the long run and it's super hard to maintain.

What I do today is simply run against the specific test folder. eg.

```
Invoke-Maester ./tests/Maester/Intune/
```

Another feature that I do want to add to Maester is to show the product icon in the test results. So if it's an Entra test we show the Entra icon etc. for that I definitely want to add a product metadata to each test. 

My thinking is to add this info to maester-config.json and render that in the UX, and of course there will be a way to add this using tags for custom tests, but I do want to avoid overloading more info into the Pester tags.

Long term there may be a future where we don't use Pester so I want to avoid adding more dependencies that require Pester.

merill:
@kij got it, so this is specifically when we run just one test or a group and they are all skipped because we are not connected. 

However I do feel the ones with Error should be a top level count in the header (if there is an error).

I like the idea of doing a hover on Not Tested that breaks out the count by the different sub categories of why it is under not tested (Skipped reason etc..).

We can then also add a sub filter for it...

How do we want to progress with this PR?

PR #1134, opened 2025-08-29
Prevent spf infinite recursion due to self-referencing redirect directive

mattcave:
When running the MS.EXO.2.2 (SPF) test against a domain that has a recursive loop in the SPF records (i.e. it redirects to itself), the test fails to identify the misconfiguration and goes into an infinite loop, ultimately resulting in a stack overflow. This results in the execution aborting and other tests not being run.

This PR adds a simple check in the Resolve-SPFRecord function to prevent this.

This change is very similar to the change in https://github.com/maester365/maester/pull/1070 but handles misconfigured *redirect* directives, rather than *include* directives.


merill (approved):
Good fix. Thanks for catching this and the PR @mattcave

Issue #994, opened 2025-06-25
Track and document tags


svrooij:
> I added the -Full tag, since this test can take a long time in enterprise tenants with 100k> apps.
>_Originally posted by @merill in https://github.com/maester365/maester/issues/912#issuecomment-2869411288_

I say we should document some of the tags somewhere. And maybe think of this problem the other way around. Meaning run this test by default and have people exclude a special tags we then give to tests that are known to fail for large tenants

SamErde:
@svrooij, can you please update the title of this issue to reflect the 'Track tags' topic? 🙏🙂

SamErde:
Related questions:

- What is the difference between the `All` tag and the `Full` tag? Does?
  - `All` = "all normal tests"
  - `Full` = "All + the tests that people don't normally include, such as the CA WhatIf"
- Should all tests have the `Full` tag and most tests have the `All` tag?
- Should it be assumed (and validated by Pester tests) that:
  - All tests in the `Maester` folder have the `Maester` tag
  - All tests in the `CIS` folder have the `CIS` tag
  - ...and so on?

@merill, @f-bader, @Cloud-Architekt

merill:
So here's the difference. 

I added Full for the tests that take a long time to run. Eg customers with 50k plus users will result in 5+ or 10+ hours to run. 

These tests require all service principals to be downloaded which can be in millions in larger tenants. 

The All tag was for tests that didn't take much time like the what if but the apis for it were in preview and not stable. 

So there is a need to exclusively flag and run these types of tests. 

Maybe we can have better names to flag these two types.

merill:
NoLimit doesn't convey the meaning. This is what Gemini proposed. I like LongRunning, that way the user is saying to include the LongRunning tests...

Category | Suggested Name | Rationale
-- | -- | --
Focus on duration | `LongRunning` | This is the most direct and universally understood term in software development for this purpose.
Focus on Thoroughness | `Comprehensive` | This name frames the long duration in a positive light. It implies the tests are slow because they are exhaustive and provide deep validation, which is exactly your scenario.
Focus on Scale | `LargeScale` | This is a great choice because it describes why the tests are slow—they are designed to run against large-scale environments (like tenants with 50k+ users). It sets the context correctly.
Focus on Time | `TimeIntensive` | A slightly more formal and descriptive alternative to LongRunning. It clearly communicates that a significant time investment is required to run the test.
[...]

PR #1138, opened 2025-08-30
Added "Full" Tag to multiple Tests

l-gosling:
### Description

I run Maester with the "Full" Tag and not all Tests are executed. This PR adds the Full Tag. For all of them exclude the CA What-If tests.


SamErde:
Thanks for this effort! There has been an ongoing conversation around the meaning and purpose of tags, and changes will be rolled out in September. I will be drafting a blog post soon to help prepare admins with automated Maester runs (and documentation) for the changes described in [this discussion](https://github.com/maester365/maester/issues/994#issuecomment-3069861772).

In short, the `All` and `Full` tags are being deprecated. In most cases, they are no longer needed, and in cases where they _are_ relevant, we are changing to more intuitive tags (`Preview` and `LongRunning`). The expected behavior will be that all tests will run automatically for the connected services _except_ the `LongRunning` tags. Those will only run if specifically requested. In the future, we also hope to be able to determine threshholds for tenant size so even long-running tests (affected by very large numbers of users and groups) will run automatically as long as the tenant size is below relevant thresholds.

l-gosling:
Thanks for the explanation, Sam!
I look forward to your post.

PR #1139, opened 2025-08-30
FIx: Test failed when no Subsciptions found. Not get skipped with reason

l-gosling:
### Description

Moved the get tenant root group to a function and fixed the bug using Add-MtTestResultDetail for this test.

### DIFF
--- a/powershell/public/maester/azure/Get-AzureRootManagementGroup.ps1
+++ b/powershell/public/maester/azure/Get-AzureRootManagementGroup.ps1
@@ -1,0 +1,42 @@
+﻿<#
+.SYNOPSIS
+    Gets the Azure Tenant Root Management Group ID
+
+.DESCRIPTION
+    This function retrieves the Tenant Root Management Group ID by querying all management groups
+    and finding the one with the display name "Tenant Root Group".
+
+.EXAMPLE
+    Test-AzureRootManagementGroup
+
+    Returns the ID of the Tenant Root Management Group if found, otherwise returns $null.
+
+.LINK
+    https://maester.dev/docs/commands/Test-AzureRootManagementGroup
+#>
+function Get-AzureRootManagementGroup {
+    [CmdletBinding()]
+    [OutputType([string])]
+    param()
+
+    try {
+        $mgResponse = Invoke-MtAzureRequest `
+            -RelativeUri "/providers/Microsoft.Management/managementGroups" `
+            -ApiVersion "2020-05-01"
+
+        $mgList = $mgResponse.value
+    }
+    catch {
+        Write-Verbose "Error retrieving management groups: $_"
+        return $null
+    }
+
+    $rootGroup = $mgList | Where-Object { $_.properties.displayName -eq "Tenant Root Group" }
+
+    if (-not $rootGroup) {
+        Write-Verbose "Tenant Root Group not found in management groups."
+        return $null
+    }
+
+    return $rootGroup.name
+}
--- a/powershell/public/maester/azure/Test-MtManagementGroupWriteRequirement.ps1
+++ b/powershell/public/maester/azure/Test-MtManagementGroupWriteRequirement.ps1
@@ -24,27 +24,13 @@ function Test-MtManagementGroupWriteRequirement {
         return $null
     }
 
-    try {
-        $mgResponse = Invoke-MtAzureRequest `
-            -RelativeUri "/providers/Microsoft.Management/managementGroups" `
-            -ApiVersion "2020-05-01"
+    $rootMgId = Get-AzureRootManagementGroup
 
-        $mgList = $mgResponse.value
-    }
-    catch {
-        Add-MtTestResultDetail -SkippedBecause Error -SkippedError $_
-        return $null
-    }
-
-    $rootGroup = $mgList | Where-Object { $_.properties.displayName -eq "Tenant Root Group" }
-
-    if (-not $rootGroup) {
+    if (!($rootMgId)) {
         Write-Verbose "Tenant Root Group not found in management groups."
-        Add-MtTestResultDetail -SkippedBecause "Tenant Root Group not found"
+        Add-MtTestResultDetail -SkippedBecause "Custom" -SkippedCustomReason "Tenant Root Group not found"
         return $null
     }
-
-    $rootMgId = $rootGroup.name
 
     try {
         $settingResponse = Invoke-MtAzureRequest `


l-gosling:
Before:
image

After:
image

powershell/public/maester/azure/Get-AzureRootManagementGroup.ps1
```diff
@@ -0,0 +1,42 @@
...
+            -ApiVersion "2020-05-01"
+
+        $mgList = $mgResponse.value
+    }
+    catch {
+        Write-Verbose "Error retrieving management groups: $_"
+        return $null
+    }
+
+    $rootGroup = $mgList | Where-Object { $_.properties.displayName -eq "Tenant Root Group" }
```
SamErde: The tenant root group displayName can be changed, and I have seen advise in some places that encourages people to set their own displayName for it. What I learned in reviewing this is that the name (GUID) of the tenant root group always matches the tenant ID. Given that, it looks like the most reliable check would be to look for a management group that has a `name` attribute that matches the tenant ID or the `id` attribute matches `/providers/Microsoft.Management/managementGroups/{tenantId}`.

powershell/public/maester/azure/Get-AzureRootManagementGroup.ps1
```diff
@@ -0,0 +1,42 @@
...
+    [CmdletBinding()]
+    [OutputType([string])]
+    param()
+
+    try {
+        $mgResponse = Invoke-MtAzureRequest `
+            -RelativeUri "/providers/Microsoft.Management/managementGroups" `
+            -ApiVersion "2020-05-01"
+
+        $mgList = $mgResponse.value
```
SamErde: Minor: You could reduce the # of saved variables by selecting the value attribute directly in the first call, like: 

```powershell
        $ManagementGroups = Invoke-MtAzureRequest `
            -RelativeUri "/providers/Microsoft.Management/managementGroups" `
            -ApiVersion "2020-05-01" | Select-Object value
```

(You _could_ even query the management group name directly, I think.)

SamErde (changes requested):
Nice solution! This just needs one simple change to accurately detect the tenant root group if/when the display name of the root MG has been changed.

SamErde:
I think your check for management groups is valuable and should stay, but have a couple of additional thoughts around this test:

- You could change the new function to list all management groups and then update the test itself to request the root management group. This function would then be usable by any other tests that might inspect management groups in general.
- Could this test also be skipped if the tenant has no Azure subscription (e.g., the title of the related issue)?

Thanks!

powershell/public/maester/azure/Get-AzureRootManagementGroup.ps1
l-gosling: Really nice!!
I search for that but not finde that information.

powershell/public/maester/azure/Get-AzureRootManagementGroup.ps1
l-gosling: Changed to:
        $ManagementGroups = Invoke-MtAzureRequest `
            -RelativeUri "/providers/Microsoft.Management/managementGroups" `
            -ApiVersion "2020-05-01" | 
            Select-Object -ExpandProperty value
            
            Without "-ExpandProperty" it the same output then before.

l-gosling:
> I think your check for management groups is valuable and should stay, but have a couple of additional thoughts around this test:
> 
> * You could change the new function to list all management groups and then update the test itself to request the root management group. This function would then be usable by any other tests that might inspect management groups in general.
> * Could this test also be skipped if the tenant has no Azure subscription (e.g., the title of the related issue)?
> 
> Thanks!

1. Yes
2. Second: You want the following block in the "Get-AzureRootManagementGroup" function too?
```
    if (!(Test-MtConnection Azure)) {
        Add-MtTestResultDetail -SkippedBecause NotConnectedAzure
        return $null
    }
```

If i look at another like "Get-MtUserAuthenticationMethod.ps1" i can't find something like that.

SamErde:
> 2. Second: You want the following block in the "Get-AzureRootManagementGroup" function too?
> 
> ```
>     if (!(Test-MtConnection Azure)) {
>         Add-MtTestResultDetail -SkippedBecause NotConnectedAzure
>         return $null
>     }
> ```
> 
> If i look at another like "Get-MtUserAuthenticationMethod.ps1" i can't find something like that.

I don't think you need that in `Get-AzureManagementGroup` because `Add-MtTestResultDetail` is performed by the Pester test script itself and not in the helper functions.

powershell/public/maester/azure/Get-AzureManagementGroup.ps1
SamErde, line 1: My apologizes, but some ideas and insights are coming to me in "real time." 😅 It would be good to just return the original objects captured from `value`. That will allow other functions to query management groups by Id, Name, DisplayName, or whatever other combination of attributes is useful for each use case. Returning the whole object here (instead of just the name) will create more flexibility, but of course, will create slightly more work in the functions/tests that call Get-AzureManagementGroup.

powershell/public/maester/azure/Test-MtManagementGroupWriteRequirement.ps1
```diff
@@ -24,38 +24,32 @@ function Test-MtManagementGroupWriteRequirement {
...
 
-        $mgList = $mgResponse.value
-    }
-    catch {
-        Add-MtTestResultDetail -SkippedBecause Error -SkippedError $_
-        return $null
-    }
+    # Find the root management group (matches tenant ID)
+    $tenantId = (Get-MgContext).TenantId
+    $rootManagementGroup = $managementGroups | Where-Object { $_ -eq $tenantId }
```
SamErde: Right on. Now, if you care to adopt my most recent suggestion (return the whole management group object[s]), you'll just need to reference the relevant attribute in this filter: `$ManagementGroups | Where-Object {$_.id -match "$tenantId$"}`. (The $ at the end makes sure the **id** _ends with_ the tenant ID.)

SamErde (changes requested):
Awesome. Very close!

powershell/public/maester/azure/Test-MtManagementGroupWriteRequirement.ps1
l-gosling: I tested a little bit. the names are unique. Use a -eq as operator hat the same result. Is this ok for you too?

powershell/public/maester/azure/Get-AzureManagementGroup.ps1
l-gosling: done

powershell/public/maester/azure/Test-MtManagementGroupWriteRequirement.ps1
SamErde: If the whole object is being returned, you will still need to specify an attribute to compare vs just referencing `$_`. That's why I provided the above example. It's possible that I may have overlooked something simple, though! Keep in mind that the Id attribute includes the whole path (and the root MG ID ends with the tenant ID).

l-gosling: in the commit i changed it back to the whole object i use the following:
`$tenantId = (Get-MgContext).TenantId
    $rootManagementGroup = $managementGroups | Where-Object { $_.name -eq $tenantId }`

powershell/public/maester/azure/Get-AzureManagementGroup.ps1
SamErde, line 1: This looks good. @merill, should every function have the Mt prefix or only certain ones?

powershell/public/maester/azure/Test-MtManagementGroupWriteRequirement.ps1
```diff
@@ -24,38 +24,32 @@ function Test-MtManagementGroupWriteRequirement {
...
 
-        $mgList = $mgResponse.value
-    }
-    catch {
-        Add-MtTestResultDetail -SkippedBecause Error -SkippedError $_
-        return $null
-    }
+    # Find the root management group (matches tenant ID)
+    $tenantId = (Get-MgContext).TenantId
+    $rootManagementGroup = $managementGroups | Where-Object { $_.name -eq $tenantId }
```
SamErde: We can avoid the additional variable and call to Get-MgContext by pulling the tenant ID from the management group object itself. For example,

`$managementGroups | Where-Object { $_.name -eq $_.properties.tenantid }`

powershell/public/maester/azure/Test-MtManagementGroupWriteRequirement.ps1
```diff
@@ -24,38 +24,32 @@ function Test-MtManagementGroupWriteRequirement {
...
         return $null
     }
 
-    $rootMgId = $rootGroup.name
-
     try {
+        # Query the management group settings to check authorization requirements
         $settingResponse = Invoke-MtAzureRequest `
-            -RelativeUri "/providers/Microsoft.Management/managementGroups/$rootMgId/settings/default" `
+            -RelativeUri "/providers/Microsoft.Management/managementGroups/$rootManagementGroup/settings/default" `
```
SamErde: I think $rootManagementGroup needs to be `$($rootManagementGroup.name)` now after all of the changes.

powershell/public/maester/azure/Test-MtManagementGroupWriteRequirement.ps1
l-gosling: Great catch! Then it works still when no graph connection is established and don't need an extra call :) 
Is changed

powershell/public/maester/azure/Test-MtManagementGroupWriteRequirement.ps1
l-gosling: yes. solved.

l-gosling:
Initial bug is fixed. 
New function to get Management group is created. Get-AzureManagementGroup.ps1
To get the tenant root group switch from displayname to id. The "name" attribute of the tenant root management is the tenant id.
image

Now waiting for the info if the function needs "Mt" in its name.

powershell/public/maester/azure/Get-AzureManagementGroup.ps1
merill: We ideally want to have the Mt prefix to prevent clashes with other modules.

powershell/public/maester/azure/Get-AzureManagementGroup.ps1
```diff
@@ -0,0 +1,35 @@
...
+    Get-AzureManagementGroup
+
+    Returns all management groups in the tenant.
+
+.LINK
+    https://maester.dev/docs/commands/Get-AzureManagementGroup
+#>
+function Get-AzureManagementGroup {
+    [CmdletBinding()]
+    [OutputType([string])]
```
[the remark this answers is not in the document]
SamErde: Makes sense. The [string] output type was relevant in the first version of this script that returned only the name.

powershell/public/maester/azure/Test-MtManagementGroupWriteRequirement.ps1
```diff
@@ -24,38 +24,28 @@ function Test-MtManagementGroupWriteRequirement {
...
-            -ApiVersion "2020-05-01"
-
-        $mgList = $mgResponse.value
-    }
-    catch {
-        Add-MtTestResultDetail -SkippedBecause Error -SkippedError $_
-        return $null
-    }
+    # Get all management groups in the tenant and filter the tenent root management group by id
+    $rootManagementGroup = Get-AzureManagementGroup | Where-Object { $_.id -match "$($_.properties.tenantid)$" }
```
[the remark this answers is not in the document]
SamErde: Current version is more precise than the Copilot suggestion. No change needed.

l-gosling:
@merill thanks for fixing the final stuff.

PR #1195, opened 2025-09-15
💫feat: add Save-MaesterOffline function for offline module installation

SamErde:
### Description

Add `Save-MaesterOffline.ps1` function to download Maester and module dependencies locally so they can be copied and installed on a server that does not have access to the PowerShell Gallery.

- Supports both PSResourceGet and legacy PowerShellGet modules
- Includes optional ZIP creation for easy distribution
- Handles prerelease modules and version specifications
- Provides comprehensive error handling and user feedback"


merill (approved):
Love it! Would be good to include some notes in the doc about this feature.

Issue #697, opened 2025-02-24
#️⃣ Pick next Maester test sequence number


merill:
This issue tracks the next available sequence number for a new core Maester test (tests starting with **MT._xxxx_** prefix)

If you are starting work on a new test

* Look up the last number that's been reserved
* Add a new comment with number you are reserving
* Start writing your new test knowing that no one else can _steal_ 😂 your test number

Note: This sequence number does not apply to tests from CISA, CIS, ORCA and EIDSCA as they already provide unique #s for their tests.

merill:
MT.1052 - Reserved for CA policy on DCF

merill:
MT.1055 - Microsoft 365 Teams/Group creation restricted to approved users

Oppedijk:
MT.1056 - Azure User Access Admin should not be enabled

svrooij:
`MT.1057` - App registrations should not have secrets

l-gosling:
MT.1058 - Exchange Application Access Policies must be configured

merill:
MT.1059 - Microsoft Defender for Identity @Cloud-Architekt fyi

svrooij:
`MT.1060` - (experimental) drift tests using file detection.

RobbeVandenDaele:
MT.1061 Mfa on device registration conflict

marcel-ngn:
MT.1063 -  All Application Owners should have MFA set up

brianveldman:
MT.1064 - Ensure that write permissions are required to create new management groups

SamErde:
MT.1066 - Ensure that conditional access policies do not reference invalid or missing users, groups, or roles

marcel-ngn:
MT.1068 - Restrict non-admin users from creating tenants

marcel-ngn:
MT.1069 - Restrict non-admin users from creating security groups

jflieben:
MT.1071 - Require explicit assignment of Third Party Entra Apps

SamErde:
> MT.1073 - Require explicit assignment of Third Party Entra Apps
> 
> edit: changed from 1071 to 1073 as it seems 1071 and 1072 are already in use as per maester-config.json :(

Thanks for taking the time to add your test # here. Unfortunately, there's also a pending PR for [1073](https://github.com/maester365/maester/pull/1108). Could you please bump yours to 1074?

I think we'll need to edit the PR template to make it harder for others to miss this step. 😉

Cloud-Architekt:
[...]

PR #1208, opened 2025-09-17
Cisa.ms.exo.10.2 emails identified as containing malware shall be quarantined or dropped. fix bugs

JeanPhilippeGeorge:
### Description

A lot of the tests concerning malware, phishing, spam, safe attachments, safe links, ... are failing on my tenants because they are not using standard and / or strict policies. That alone make the tests fail.

And in the case standard and strict policies exist but are assigned to no user it's possible to pass the tests with a custom policy with the property QuarantineTag set to anything due to this: 
```ps
$policies = Get-MtExo -Request MalwareFilterPolicy

$fileFilter = $policies | Where-Object { `
    $_.QuarantineTag -eq "AdminOnlyAccessPolicy"
}
# if standard and strict exists the following is always true
$testResult = $standard -and $strict -and (($fileFilter|Measure-Object).Count -ge 1)
```

I think that no policies should have QuarantineTag set to anything else that "AdminOnlyAccessPolicy" and I have changed the code to this:
```ps
    $policies = Get-MtExo -Request MalwareFilterPolicy
    $failingPolicies = $policies | Where-Object { $_.QuarantineTag -ne "AdminOnlyAccessPolicy" }

    $testResult = ($failingPolicies | Measure-Object).Count -eq 0
```

The direct link from Scubagear for this CISA test doesn't mention explicitly standard/strict policies: https://github.com/cisagov/ScubaGear/blob/main/PowerShell/ScubaGear/baselines/exo.md#msexo102v1, but it's mentionned a lot of time before and after.

I think it would be way better to have a different (new) test that really tests that everybody is assigned to a standard or strict policy. 

And individual tests that are testing what they say they are testing.

I could add PR fot the other tests if it's ok for you?

The doc, messages would also need to be corrected. I can make a first pass, but a review would be necessary to correct my english ;-)

### Your checklist for this pull request

🚨Please review the [guidelines for contributing](https://maester.dev/docs/contributing) to this repository.

- [x] Read the guidelines for contributions

#### PRs related to all code

- [x] Before you submit the PR, run the tests locally by running `/powershell/tests/pester.ps1`
- [ ] After submitting, verify the tests are still passing on your PR in GitHub (the tests are run across all platforms).

#### PRs related to Maester Tests

If you are contributing a test or making updates to a test please verify these.

- [ ] Check that the unique ID you want to use is not already picked. See [Pick next Maester test sequence number](https://github.com/maester365/maester/issues/697)
- [ ] Add Test with your .ps1 and .md-file (make sure .ps1 file is encoded with "UTF8 with BOM") `.\powershell\public\maester...`
- [ ] Add your new test-function to a corresponding .test.ps1 file `\tests\Maester..*.test.ps1`
- [ ] Add your new test-function to 'FunctionsToExport'-List `.\powershell\Maester.psd1`
- [ ] Add test Id, Severity and Title to "maester-config.json" file `.\tests\maester-config.json`
- [ ] Add website documentation for test with name of "testId.md" `.\website\docs\tests\maester`
- [ ] If Invoke-Maester parameters are changed, update the GitHub action to use the new parameters.


merill:
@JeanPhilippeGeorge Tx for the PR. I like your idea of splitting out the assignments to another test. This will help customers pass at least some tests if they have the policy and another one that checks assignments. 

We eventually need to do this for CA policies, Intune config policies etc.. so it's a good start. 

The only thing we need to work out is whether these go under the CISA umbrella (if CISA doesn't enforce in their checks) or we add them as Maester tests. If CISA does not require it then I think it should be part of the Maester tests.

Now for this specific PR, I guess my question is, does this align with CISA ScubaGear? We want the Maester versions of the tests to be in sync with CISA ScubaGear tests. If they faily in CISA they should fail here and vice versa.

Thoughts @soulemike

JeanPhilippeGeorge:
@merill thanks for your thanks and thank you and rest of the team for a great project 😊

When I created this PR I thought it would be controversial 😉

When I started using Orca and Scubagear it was great: I learned a ton of thing and my tenants became more secure.

So, a huge thanks to Orca and Scubagear team for their work!

But there were a lot of bugs, and the project seemed to be dead.

When I first heard of Maester it seemed great as it was and still is very active. But these days it seems to have a lot of bugs too which makes it harder to work with.

> The only thing we need to work out is whether these go under the CISA umbrella (if CISA doesn't enforce in their checks) or we add them as Maester tests. If CISA does not require it then I think it should be part of the Maester tests.
>
> Now for this specific PR, I guess my question is, does this align with CISA ScubaGear? We want the Maester versions of the tests to be in sync with CISA ScubaGear tests. If they faily in CISA they should fail here and vice versa.
> 

The problem here is that Scubagear tests are nor testing what they say they are testing. 

**I changed my config to this:**

image


image

  
**And now test Cisa.ms.exo.10.2 succeeds:**

image

**My suggestion correctly fails:**

image
 
**So here are some choices:**
+ Do exactly the same as Scubagear and let users mistakenly think they are secure.
+ At least fix the test by making it fail if any policy has the wrong setting
+ Fix the test (and the other that have the same problem) and add a test for strict / standard policy enabled which actually tests the state and the fact that users are assigned to the policies

Scubagear CISA doc: https://github.com/cisagov/ScubaGear/blob/main/PowerShell/ScubaGear/baselines/exo.md#10-malware-scanning 

image

So, under 10.Malware scanning: it’s required to enable standard/strict policy

But under10.2.v1(and friends), there is no mention of presets. Only that malware shall be dropped.

I think it would be preferable to respect the spirit of the law here. But maybe I’m biased 😉
 

I understand that it is not an easy decision.

Note: to test the real state of strict and standard presets:
```ps
Get-ATPProtectionPolicyRule | select identity, state

Identity                        State
--------                        -----
Standard Preset Security Policy Disabled
Strict Preset Security Policy   Disabled

Get-EOPProtectionPolicyRule | select identity, state

Identity                        State
--------                        -----
Standard Preset Security Policy Disabled
Strict Preset Security Policy   Disabled
```

powershell/public/cisa/exchange/Test-MtCisaMalwareAction.ps1
```diff
@@ -21,51 +21,24 @@ function Test-MtCisaMalwareAction {
     if (!(Test-MtConnection ExchangeOnline)) {
         Add-MtTestResultDetail -SkippedBecause NotConnectedExchange
         return $null
-    } elseif (!(Test-MtConnection SecurityCompliance)) {
```
soulemike, line 24: Due to the current Get-MtExo dependency, I believe removing this could cause errors. If tested then no concerns, but just a thought that came to mind. It does look like that cmdlet is from the Exo Module and not Ipp, but more just concerned how the cache handling will occur.

JeanPhilippeGeorge: I tested without Connect-IPPSSession and the test is executed and give the same result than before.

Not competent for the cache handling part.

merill:
@JeanPhilippeGeorge would you be able to answer @soulemike so we can work out how we move forward?

With the CISA tests, many gov agencies may be using Maester and they would expect the results to be in line with the CISA ScubaGear results. 

We can however move this into it's own test in Maester. We would love it if you can move your version of the test into an independent Maester test. Tx!

JeanPhilippeGeorge:
Sorry to have taken that much time to answer, that was a tough week!

@soulemike 

+ Is the CISA test not functioning according to the test description?

It's not functionning, see my previous comments

+ Is there a separate issue for an ORCA test?

Not from me.

+ Can you move this test to a new MT test as this additional capability would be a valuable addition?

I can, but some other tests in the CISA.MS.EXO tests (those which test the threat policies) seem to have the same problem: if strict/standard policies exists the test pass, there is no need to enable them ; if there is no strict / standard policies the test fail even if custom policy with the correct setting exists. 

So, it’s not just that test, it’s a set of tests.

I’m not confident for CISA.MS.EXO.9.xx to CISA.MS.EXO.11.xx, and CISA.MS.EXO.14.xx to CISA.MS.EXO.15.xx

I have not looked at the code of all this tests, only some.

@merill
I can make a new test for 'CISA.MS.EXO.10.2: Emails identified as containing malware SHALL be quarantined or dropped' as a start. But the other test will need to be done too ...

I would show test results like that:

|Policy enabled |Policy name| QuarantineTag| Result |
|--|--|--|--|
| false |Standard|SomethingElse| skipped|
|true|Strict|AdminOnlyAccessPolicy|passed|
|true| Custom |AdminOnlyAccessPolicy|passed|

The test pass because all enabled policies have QuarantineTag set to AdminOnlyAccessPolicy

I can take the test detail from the CISA test and add the Powershell commandlet to correct the parameter.

Would that be ok for you?

If it's ok, we can close that PR and I will create a new one.

I have no idea how to name this test (and the others), where to put it (tests\Maester\Exchange or tests\Maester\Exchange\CISA), category, tag, …

powershell/public/cisa/exchange/Test-MtCisaMalwareAction.ps1
JeanPhilippeGeorge: @soulemike thoughts on my answer?

What should I do:
+ Revert?
+ Keep it like that and resolve conversation?

merill: Let's keep in and review to see how it goes with the preview build.

PR #1218, opened 2025-09-24
Shorten, standardize Exposure Management test describe value

SamErde:
# Description

Updated the `Describe` attribute of the Exposure Management test to be consistent with how all other Maester tests use the `Describe` attribute as a tag, or short label.

image


merill (approved):
Thanks for the update @SamErde

PR #1223, opened 2025-09-27
New feature - Maester DevOps App management cmdlets

merill:
Setting up Maester Automation in GitHub or Azure DevOps can take 15+ minutes and numerous clicking and switching between multiple portals. The bulk of this time is spent on creating the Entra app for the Maester automation account and adding all of the app permissions.

Its also a lot of work to figure out what permissions your Entra app doesn't have and to keep them up to date as new Maester tests are added.

With the addition of four new cmdlets life becomes much easier for setting up and maintaining Entra apps for Maester automation.

* `New-MtMaesterApp`
* `Update-MtMaesterApp`
* `Get-MtMaesterApp`
* `Add-MtMaesterAppFederatedCredential`

 Now you can simply run this command to create the Entra app and add all it's permissions.

Creating a Maester daily automation set up went from 50+ steps to less than 5 steps. You don't even need to go to the Entra portal anymore!

Here's how 👇

```
New-MtMaesterApp
```

If you are using GitHub for your Maester automation you can supply your repository details to set up Workload ID federation in one go.

```
New-MtMaesterApp -GitHubOrganization merill -GitHubRepository maester-demo
```

If you have an existing Maester automation account set up in Entra and you want to add all the latest required Graph permission run `Update-MtMaesterApp`. This is also handy to add all the required permissions to the managed identity created by Azure DevOps.

```
Update-MtMaesterApp -AppId 'your-app-id'
```

All Entra apps created or updated with these cmdlets are automatically tagged with a 'maester' tag. This helps you better manage your Maester related identities in Entra and avoid creating duplicate accounts.

Simply run `Get-MtMaesterApp` to get a list of all the Maester related Entra apps in your tenant. We even include a PortalLink so you are only one-click away from the command line to opening the app in the Entra portal.

If you have an existing app and want to add Workload ID federation (maybe to move away from client secrets) you can use `Add-MtMaesterAppFederatedCredential` for e.g.

```
Add-MtMaesterAppFederatedCredential -AppId 'your-app-id' -GitHubOrganization 'merill' -GitHubRepository 'maester-demo'
```

An important note. These three cmdlets use the Az PowerShell modules to create the apps and credentials, this is to avoid asking the user to consent to highly privileged permissions like Application.ReadWrite.All to the Graph PowerShell SDK as part of setting up Maester.

We always want to keep Maester requirements to read-only permission scopes.

Why does it work for Az PowerShell? Entra app management permissions are pre-consented for the Az PowerShell module and using these instead of Graph PS helps avoid blocking scenarios for anyone trying to set up Maester.

TODO in upcoming PRs

* Add support for adding optional permissions (Exchange, Azure, etc)
* Update monitoring docs to default to using these cmdlets (while retaining the manual steps)
* Blog post for this feature announcement (Use the text in this PR as a starting point)
* Add support for Mail.Send to use least privilege (access control so permission is only granted to the mailbox used for sending)
* Future: Add support to automate creating the GitHub repo creation so the entire daily automation set up can be run from the commandline with a single command.


powershell/internal/Get-MaesterAppInfo.ps1
```diff
@@ -0,0 +1,27 @@
+﻿function Get-MaesterAppInfo {
```
SamErde: **Nit picks:** Functionally looks good. It's pretty self-explanatory but including at least a minimal comment-based help description is always nice. The formatting style doesn't seem consistent with the rest of the project from what I recall.

powershell/internal/Test-MtAzContext.ps1
```diff
@@ -0,0 +1,22 @@
...
+    This cmdlet is used mainly by the Get-MtMaesterApp and Update-MtMaesterApp cmdlets and related cmdlets.
+#>
+
+function Test-MtAzContext {
+    [CmdletBinding()]
+    [OutputType([bool])]
+    [Diagnostics.CodeAnalysis.SuppressMessageAttribute('PSAvoidUsingWriteHost', '', Justification = 'Colors are beautiful')]
+    param ()
+
+    $validContext = $true
```
SamErde: Should this be swapped to start without a valid context and only set it to $true if the context is confirmed?
Are any permissions required?

powershell/internal/Set-MaesterAppPermissions.ps1
```diff
@@ -0,0 +1,92 @@
...
+    process {
+        Write-Verbose "Setting permissions for app: $AppId"
+        Write-Verbose "Scopes: $($Scopes -join ', ')"
+
+        # Get the service principal for the app
+        $appSPResponse = Invoke-MtAzureRequest -RelativeUri "servicePrincipals" -Filter "appId eq '$AppId'" -Method GET -Graph
+        $appSPs = $appSPResponse.value
+
+        if ($appSPs.Count -eq 0) {
+            throw "Service principal for application not found. Ensure the application has a service principal."
```
SamErde: If the app ID is not found, this `throw` will end the function immediately. It might be helpful to still allow the Graph application to get checked before ending so admins can see if that SP might be usable.

powershell/internal/Set-MaesterAppPermissions.ps1
```diff
@@ -0,0 +1,92 @@
...
+        # Process each requested scope
+        foreach ($scope in $Scopes) {
+            Write-Verbose "Processing scope: $scope"
+
+            # Find the app role for this scope
+            $appRole = $graphAppRoles | Where-Object { $_.value -eq $scope }
+
+            if (-not $appRole) {
+                Write-Warning "Application permission '$scope' not found in Microsoft Graph. Skipping."
+                continue
```
SamErde: This `continue` is unnecessary.

powershell/internal/Set-MaesterAppPermissions.ps1
```diff
@@ -0,0 +1,92 @@
...
+            }
+
+            # Check if permission already exists
+            $existingPermission = $currentAppRoles | Where-Object { $_.appRoleId -eq $appRole.id -and $_.resourceId -eq $graphSP.id }
+
+            if ($existingPermission) {
+                Write-Verbose "Permission '$scope' already exists"
+                continue
```
SamErde: This `continue` is unnecessary.

powershell/internal/Set-MaesterAppPermissions.ps1
```diff
@@ -0,0 +1,92 @@
...
+function Set-MaesterAppPermission {
+    [Diagnostics.CodeAnalysis.SuppressMessageAttribute('PSAvoidUsingWriteHost', '', Justification = 'Colors are beautiful')]
+    [Diagnostics.CodeAnalysis.SuppressMessageAttribute('PSUseShouldProcessForStateChangingFunctions', '', Justification = 'TODO: Implement ShouldProcess')]
+    [CmdletBinding()]
+    param(
+        [Parameter(Mandatory = $true)]
+        [string] $AppId,
+
+        [Parameter(Mandatory = $true)]
+        [string[]] $Scopes
```
SamErde, line 27: It would be awesome if we could offer "templates" of scopes to apply to provide the minimum necessary permissions for things like:

- all checks
- Defender checks
- PIM checks (extended privs required)
- send email report
- send Teams report

etc

powershell/internal/Set-MaesterAppPermissions.ps1
```diff
@@ -0,0 +1,92 @@
...
+            Write-Host "➕ Adding permission '$scope'..." -ForegroundColor Yellow
+            # Add the permission
+            $permissionBody = @{
+                principalId = $appSP.id
+                resourceId  = $graphSP.id
+                appRoleId   = $appRole.id
+            } | ConvertTo-Json
+
+            Write-Verbose "Adding permission: $scope"
+            Invoke-MtAzureRequest -RelativeUri "servicePrincipals/$($appSP.id)/appRoleAssignments" -Method POST -Payload $permissionBody -Graph | Out-Null
```
SamErde, line 86: A try/catch would be helpful if output is piped to null.

powershell/public/core/Add-MtMaesterAppFederatedCredential.ps1
```diff
@@ -0,0 +1,171 @@
...
+        [Parameter(Mandatory = $false)]
+        [string] $GitHubBranch = 'main',
+
+        # The name for the federated credential
+        [Parameter(Mandatory = $false)]
+        [string] $Name
+    )
+
+    if (-not (Test-MtAzContext)) {
+        return
```
SamErde, line 76: Missing output to the user?

powershell/public/core/Add-MtMaesterAppFederatedCredential.ps1
```diff
@@ -0,0 +1,171 @@
...
+    try {
+        if ($Id) {
+            $params = @{ Id = $Id }
+        } elseif ($AppId) {
+            $params = @{ AppId = $AppId }
+        }
+
+        $app = Get-MtMaesterApp @params
+        if (-not $app) {
+            return
```
SamErde, line 90: I love this dynamic splatting approach!

The return statement needs a warning or error for the user.

powershell/public/core/Add-MtMaesterAppFederatedCredential.ps1
```diff
@@ -0,0 +1,171 @@
...
+        Write-Host "   Name: AZURE_TENANT_ID" -ForegroundColor Cyan
+        Write-Host "    Value: $((Get-AzContext).Tenant.Id)" -ForegroundColor Cyan
+        Write-Host ""
+        Write-Host "See https://maester.dev/docs/monitoring/github#add-entra-tenant-info-to-github-repos for details." -ForegroundColor Yellow
+        Write-Host ""
+
+        return $createdCredential
+
+    } catch {
+        Write-Error "Failed to add federated credential: $($_.Exception.Message)"
```
SamErde, line 170: Something I've been curious about for a while: can this `Write-Error` statement and the `throw` statement be combined into one `throw "Failed to add federated credential: $($_.Exception.Message)"` statement, or is there an idiomatic principal to keep both?

powershell/public/Connect-Maester.ps1
```diff
@@ -123,10 +123,16 @@
          if ($Service -contains 'Azure' -or $Service -contains 'All') {
             Write-Verbose 'Connecting to Microsoft Azure'
             try {
+               $azWarning = @()
                if ($TenantId) {
-                  Connect-AzAccount -SkipContextPopulation -UseDeviceAuthentication:$UseDeviceCode -Environment $AzureEnvironment -Tenant $TenantId
+                  Connect-AzAccount -SkipContextPopulation -UseDeviceAuthentication:$UseDeviceCode -Environment $AzureEnvironment -Tenant $TenantId -WarningAction SilentlyContinue -WarningVariable azWarning
```
SamErde, line 128: This addition looks useful. Was it added for functionality in this PR or to resolve other issue requests regarding scope errors?

powershell/internal/Test-MtAzContext.ps1
merill: Good call, changed the init to the method call directly.

powershell/internal/Get-MaesterAppInfo.ps1
merill: Added comments

powershell/internal/Set-MaesterAppPermissions.ps1
merill: There isn't much the cmdlet can do without the SP. In a typical run it shouldn't happen unless they skipped the first part.

I did update the message to guide the user on how to recover.

`Service principal for application not found. Try deleting the Maester app and recreating with New-MtMaesterApp.`

powershell/internal/Set-MaesterAppPermissions.ps1
merill: Good call. Removed.

powershell/internal/Set-MaesterAppPermissions.ps1
merill: Good call. Removed.

powershell/internal/Set-MaesterAppPermissions.ps1
merill: Good idea, so this cmdlet is an internal one. 

Get-MtGraphScope is the right place where we would want all the scope related functionality to live. 

For this cmdlet the parameter come from update-mtmaesterapp and new-mtmaes... and scopes can be specified by the switches we have similar to Invoke-MtMaester...

```
.EXAMPLE
    Update-MtMaesterApp -AppId "12345678-1234-1234-1234-123456789012" -SendMail -Privileged

    Updates the specified Maester app with mail sending and privileged capabilities.

.EXAMPLE
    Update-MtMaesterApp -AppId "12345678-1234-1234-1234-123456789012" -Scopes @("User.Read.All")

    Updates the specified Maester app with additional custom scopes.
```

powershell/public/core/Add-MtMaesterAppFederatedCredential.ps1
merill: I'm not a big fan of throw since it shows a stack trace, instead of a neat error message that Write-Error provides.

powershell/public/Connect-Maester.ps1
merill: Yeah, the Az module lists warnings when it can't connect to various Azure Subscriptions that the user has access to. Since we are not really concerned about the specific subscription and there is nothing required by the user I've hidden it by default to avoid confusing the user.

powershell/public/core/Add-MtMaesterAppFederatedCredential.ps1
merill: I learnt that from Alexandar Nikolic! Good catch on the warning/error, added a message.

powershell/public/core/Add-MtMaesterAppFederatedCredential.ps1
merill: The Test-MtAzContext handles displaying the message. Didn't want each cmdlet that calls this to duplicate the error message to the user.

powershell/public/core/Add-MtMaesterAppFederatedCredential.ps1
SamErde: That's a good reason! Thanks.

powershell/internal/Set-MaesterAppPermissions.ps1
SamErde: I was thinking something simple like this unless the failure is reported to the user in the calling function.

```powershell
try {
    #...
} catch {
    Write-Verbose "Failed to update the application role assignments. $_"
}
```

merill: Since this is an internal cmdlet, my practice is to only catch if I'm doing something specific with the error (comes from my C# days). 

With the catch above, it would swallow the actual error and not let the calling method know that an error occurred and make it assume the function completed succesfully.

PR #1224, opened 2025-09-28
Allow definition of emergency access accounts

f-bader:
This pull request introduces the ability to change the configuration of the `maester-config.json` file to provide one or multiple emergency access accounts/groups

```json
{
  "GlobalSettings": {
    "EmergencyAccessAccounts": []
  }
// [...]
```

You can use ObjectIds, User UPNs, Group Names

### DIFF
--- a/build/aitools/test-metadata/maester-config.json
+++ b/build/aitools/test-metadata/maester-config.json
@@ -1,4 +1,7 @@
 {
+  "GlobalSettings": {
+    "EmergencyAccessAccounts": []
+  },
   "TestSettings": [
     {
       "Id": "CIS.M365.1.1.1",
--- a/powershell/internal/Get-MtMaesterConfigGlobalSetting.ps1
+++ b/powershell/internal/Get-MtMaesterConfigGlobalSetting.ps1
@@ -1,0 +1,32 @@
+﻿<#
+.SYNOPSIS
+    Gets the global settings from the Maester config.
+.DESCRIPTION
+    This function retrieves the global settings from the Maester config.
+    It returns the settings as a hashtable, which can be used to customize the behavior of the tests.
+.EXAMPLE
+    $globalSettings = Get-MtMaesterConfigGlobalSetting -SettingName 'EmergencyAccessAccounts'
+    # This will return the global settings for the setting with name 'EmergencyAccessAccounts'.
+#>
+
+function Get-MtMaesterConfigGlobalSetting {
+    [CmdletBinding()]
+    [OutputType([hashtable])]
+    param(
+        # The setting name of the configuration for which to retrieve the settings.
+        [Parameter(Mandatory = $true, Position = 0)]
+        [string]$SettingName
+    )
+
+    # Check if the Maester config is loaded
+    if (-not ($__MtSession -and $__MtSession.MaesterConfig -and $__MtSession.MaesterConfig.GlobalSettings )) {
+        Write-Verbose "Maester global config not loaded. Please run Get-MtMaesterConfig first to load the config."
+        return $null
+    } else {
+        Write-Verbose "Maester global config loaded"
+        Write-Verbose "Maester global config `"$SettingName`": $($__MtSession.MaesterConfig.GlobalSettings.$SettingName | ConvertTo-Json -Depth 5 -Compress)"
+    }
+
+    # Retrieve the test settings from the Maester config
+    return $__MtSession.MaesterConfig.GlobalSettings.$SettingName
+}
--- a/powershell/public/maester/entra/Test-MtCaEmergencyAccessExists.ps1
+++ b/powershell/public/maester/entra/Test-MtCaEmergencyAccessExists.ps1
@@ -26,6 +26,8 @@ function Test-MtCaEmergencyAccessExists {
         return $null
     }
 
+    $EmergencyAccessAccounts = Get-MtMaesterConfigGlobalSetting -SettingName 'EmergencyAccessAccounts'
+
     try {
         # Only check policies that are not related to authentication context (the state of policy does not have to be enabled)
         $policies = Get-MtConditionalAccessPolicy | Where-Object { -not $_.conditions.applications.includeAuthenticationContextClassReferences }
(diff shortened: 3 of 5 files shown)


powershell/internal/Get-MtMaesterConfigGlobalSetting.ps1
```diff
@@ -0,0 +1,32 @@
...
+    [OutputType([hashtable])]
+    param(
+        # The setting name of the configuration for which to retrieve the settings.
+        [Parameter(Mandatory = $true, Position = 0)]
+        [string]$SettingName
+    )
+
+    # Check if the Maester config is loaded
+    if (-not ($__MtSession -and $__MtSession.MaesterConfig -and $__MtSession.MaesterConfig.GlobalSettings )) {
+        Write-Verbose "Maester global config not loaded. Please run Get-MtMaesterConfig first to load the config."
```
SamErde, line 23: Should this logic and verbose output allow for a Maester config that does not include a GlobalSettings section? (Just considering possibilities. I don't see a problem here.)

SamErde:
I added comments/thoughts admist @merill's quoted response:
&nbsp;  

**Function Name**
> Rename Get-MtMaesterConfigGlobalSetting to something shorter like Get-MtConfigGlobalSetting -Name

I had the same thought. 😅
&nbsp;  

**Global Settings in JSON Schema**
> I feel for global settings we should use a json schema to define and validate the config (@SamErde recommended this originally). That way we can clearly document all of the global settings. We also need to come up with some guidelines on what should go into global setting (eg config that can be used in multiple tests, anything more?)

Let's loop this back into discussion #990 ([Module Configuration, Test Metadata, and JSON](https://github.com/maester365/maester/discussions/990)) Would it make sense to keep it simple and start with "Global setings that multiple tests can use (e.g. IDs of emergency access accounts)" or do you want to discuss the schema structure as a whole before implementing anything?
&nbsp;  

**Global Settings Validation**
> We definitely need some pester tests for global settings that define all the supported types. Because we want to build a UI that will allow users to pick and set values and we want to be able to do validation on these. ie what is a required value, vs optional (I believe the json schema can handle these...)

Can you describe (no pun intended) what you mean by pester tests for global settings? --unless this was a stream of consciousness in which you realized that we can use JSON schema validation instead of Pester tests for this case. 🙂
&nbsp;  

**Global Settings Object Types**
> That is for the overall thing. Fabian for this specific variable (EmergencyAccess), I like that you allow all types of entries in one parameter but looking forward to validation and a UX for config. The same logic will need to be duplicated in multiple place. I feel its better if we are explicit on the property. That way, we can also do json validation.

So what I hear is that we should distinctly define the object type as a user or group; is that correct? That does seem like it would be valuable for validation and clarity.
&nbsp;  

**JSON Schema POC**
> Here's the schema that Claude helped generate for me. What do you think?

The generated schema makes sense at first glance. I'll add it to what I was testing previously and share comments here. I haven't worked on that proof of concept since July, but here is a link to what I had started drafting for it back then: [@SamErde/feat-JSON-schema/tests/maester-config.schema.json](https://github.com/SamErde/maester/blob/feat-JSON-schema/tests/maester-config.schema.json)

f-bader:
@SamErde @merill  I finally finished the adoption of the proposed JSON schema. If I didn't miss anything this should be good for release

powershell/internal/Get-MtMaesterConfigGlobalSetting.ps1
```diff
@@ -0,0 +1,32 @@
...
+    $globalSettings = Get-MtMaesterConfigGlobalSetting -SettingName 'EmergencyAccessAccounts'
+    # This will return the global settings for the setting with name 'EmergencyAccessAccounts'.
+#>
+
+function Get-MtMaesterConfigGlobalSetting {
+    [CmdletBinding()]
+    [OutputType([hashtable])]
+    param(
+        # The setting name of the configuration for which to retrieve the settings.
+        [Parameter(Mandatory = $true, Position = 0)]
```
SamErde, line 17: Would there be any value in making this _not_ mandatory? If no **SettingName** is specified, the whole **GlobalSettings** object could be returned, otherwise the specified **SettingName** (e.g. 'EmergencyAccessAccounts') would be returned as whatever implicit type of object is stored in that setting from the JSON. I think this could be useful for discovering the global configuration settings that are available or defined. / @maester365/core-module

powershell/internal/Get-MtMaesterConfigGlobalSetting.ps1
SamErde: On a related note, `Get-MtMaesterConfig` is an internal function and not exported with the module. It won't be available to run in an interactive session.

If I'm not mistaken, `$__MtSession` and its attributes are not populated until `Invoke-Maester` is run. I'm wondering if we initialize `$__MtSession` and `$__MtSession.MaesterConfig` in the **Maester.psm1** file so it becomes an available object as soon as the module is imported. Additional session and invocation attributes could still be added during other relevant Maester functions. Any thoughts on that, @maester365/core-module?

build/aitools/test-metadata/maester-config.json
```diff
@@ -1,4 +1,20 @@
 {
+  "GlobalSettings": {
+    "EmergencyAccessAccounts": [
```
SamErde, line 3: This works great for a setup where only one tenant is assessed with Maester. For people who may run it against multiple tenants (🙋‍♂️), it introduces the problem of tenant-specific settings. Do we solve this by nesting tenants under GlobalSettings, or by separating tenant-specific settings from GlobalSettings?

At first thought, I lean towards this approach:

```json
{
  "GlobalSettings": {
    "IncludeTags": [],
    "ExcludeTags": []
  },
  "TenantSettings": {
    "12345678-1234-1234-1234-123456789012": {
      "Name": "Contoso",
      "EmergencyAccessAccounts": [
        { "Identifier": "12345678-1234-1234-1234-123456789012", "Type": "User" },
        { "Identifier": "breakglass@contoso.com", "Type": "Group" },
        { "Identifier": "emergency@contoso.com", "Type": "User" }
      ],
      "SkipTests": [
        { "Id": "CIS.M365.1.1", "Reason": "We don't test this one because..." },
        { "Id": "CISA.M365.1.1", "Reason": "We don't test this one because..." }
      ],
      "IncludeTags": [],
      "ExcludeTags": [],
      "DisableWAM": false
    },
    "12345678-1235-1234-1234-123456789012": {
      "Name": "Fabrikam",
      "EmergencyAccessAccounts": [
        { "Identifier": "12345678-1235-1234-1234-123456789012", "Type": "User" },
        { "Identifier": "breakglass@fabrikam.com", "Type": "Group" },
        { "Identifier": "emergency@fabrikam.com", "Type": "User" }
      ],
      "SkipTests": [
        { "Id": "CIS.M365.1.1", "Reason": "We don't test this one because..." }
      ],
      "IncludeTags": [],
      "ExcludeTags": [],
      "DisableWAM": true
    }
  }
}

```

What do you think? Feel free to tell me I'm over-complicating it! @maester365/core-module

powershell/public/maester/entra/Test-MtCaEmergencyAccessExists.ps1
```diff
@@ -35,48 +37,149 @@ function Test-MtCaEmergencyAccessExists {
...
+            $testResult += "These conditional access policies don't have the emergency access $EmergencyAccessUUIDType excluded:`n`n%TestResult%"
+            Add-MtTestResultDetail -GraphObjects $policiesWithoutEmergency -GraphObjectType ConditionalAccess -Result $testResult
+            return $result
+
+        } else {
+            # Resolve emergency access accounts/groups to object IDs and get display names
+            $ResolvedEmergencyAccessAccounts = @()
+            foreach ($account in $EmergencyAccessAccounts) {
+                # Use either Id or UserPrincipalName to identify the account / group and UserPrincipalName as fallback
+                $identifier = if ($account.Id) { $account.Id } else { $account.UserPrincipalName }
```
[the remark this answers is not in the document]
SamErde: Or something essentially like this for better validation of each property?

```powershell
if ([string]::IsNullOrEmpty($account.Id)) {
  Write-Warning "An empty account ID was found."
  break
} elseif ([string]::IsNullOrEmpty($account.Type) {
  Write-Warning "An account type was not specified for $($account.Id)."
  break
} else {
  continue
}
```

powershell/public/maester/entra/Test-MtCaEmergencyAccessExists.ps1
```diff
@@ -35,48 +37,149 @@ function Test-MtCaEmergencyAccessExists {
...
+                        if ($object) {
+                            Write-Verbose "Emergency access $type`: $($object.displayName) ($identifier)"
+                            $ResolvedEmergencyAccessAccounts += @{ObjectId = $object.id; displayName = $object.displayName; type = $type }
+                        } else {
+                            Write-Warning "Could not resolve emergency access $type ID: $identifier"
+                        }
+                    } catch {
+                        Write-Warning "Could not resolve emergency access $type ID: $identifier. Error: $_"
+                    }
+                } elseif ($identifier -match '^[^@]+@[^@]+\.[^@]+$') {
```
SamErde, line 104: We could leave this super-simple, or attempt full validation with a string such as: `^(?=.{1,113}$)([A-Za-z0-9._%+\-]+)@([A-Za-z0-9][A-Za-z0-9.-]*\.[A-Za-z]{2,})$`.

This AI-suggested pattern:

1. Enforces max UPN length (113 characters) where Microsoft Entra ID requires:
  - UPN prefix ≤ 64 characters
  - UPN suffix (domain) ≤ 48 characters
  - Combined limit ≈ 64 + 1 + 48 = 113 characters: `(?=.{1,113}$)`

2. A valid mailbox-style prefix where Entra usernames (the local part of the UPN) allow: `[A-Za-z0-9._%+\-]+`

  - Letters
  - Digits
  - Dot (.)
  - Hyphen (-)
  - Underscore (_)
  - Plus (+)
  - Percent (%)

3. Domain must be a valid DNS domain: `[A-Za-z0-9][A-Za-z0-9.-]*\.[A-Za-z]{2,}`

  - No starting period or hyphen
  - Must contain at least one dot
  - TLD must be at least 2 characters
  - Allows subdomains: user@mail.region.example.com

PR #1297, opened 2025-10-28
CIS remove level number from test names

tdcthosc:
# Description

To be discussed: I'd like to propose that we remove the level number from the test names to align with other test names, as we already have the tags to filter for e.g. "L1" and "CIS E3 Level 1".

## Contribution Checklist

Before submitting this PR, please confirm you have completed the following:

- [x] 📖 Read the [guidelines for contributing](https://maester.dev/docs/contributing) to this repository.
- [x] 🧪 Ensure the build and unit tests pass by running `/powershell/tests/pester.ps1` on your local system.

&nbsp;

Join us at the Maester repository [discussions](https://github.com/maester365/maester/discussions) 💬 or [Entra Discord](https://discord.maester.dev/) 🧑‍💻 for more help and conversations!


SamErde:
I'm conflicted on the idea of removing the CIS level indicator from the test title simply because the CIS benchmarks have `(L1)` or `(L2)` in their titles. I don't have _very_ strong feelings on this, though, so I'm curious what others think!

tdcthosc:
In my opinion the standalone L1 and L2 doesn't quite make sense, as when filtering for the tags it doesn't say what it's regarding.

tdcthosc:
> > In my opinion the standalone L1 and L2 doesn't quite make sense, as when filtering for the tags it doesn't say what it's regarding.
> 
> Agree 100%! I somewhat feel the same about the "Security" tag, which also feels redundant.

Agreed. Let's keep pushing and pulling! 😄

merill:
I'm actually leaning towards dropping the Ls since we have it in the metadata, plus the report also shows it in the UX so they can filter it out. In addition the L1/L2 tag is also shown in the detail page of the test.

I plan on creating stand alone pages for each of the tests in Maester and dropping the L1 will make it cleaner and help with SEO...

PR #1366, opened 2025-12-08
Initial build of AD DS tests.

soulemike:
# Description

Provides new functionality to enable Active Directory tests.

- ActiveDirectory is a new service available for Maester to connect to
- Caching of AD queries is provided through new function
- 10 initial new tests specific to Computer objects provided

image

image

## Contribution Checklist

Before submitting this PR, please confirm you have completed the following:

- [X] 📖 Read the [guidelines for contributing](https://maester.dev/docs/contributing) to this repository.
- [X] 🧪 Ensure the build and unit tests pass by running `/powershell/tests/pester.ps1` on your local system.

&nbsp;

Join us at the Maester repository [discussions](https://github.com/maester365/maester/discussions) 💬 or [Entra Discord](https://discord.maester.dev/) 🧑‍💻 for more help and conversations!

### DIFF
--- a/powershell/public/Set-MtAdCache.ps1
+++ b/powershell/public/Set-MtAdCache.ps1
@@ -1,0 +1,439 @@
+﻿<#
+.SYNOPSIS
+    Sets the local cache of AD lookups.
+
+.DESCRIPTION
+    By default all AD queries are cached and re-used for the duration of the session.
+
+    Use this function to set the cache.
+
+.PARAMETER Server
+    Server name to pass through to the AD Cmdlets
+
+.PARAMETER Credential
+    Credential object to pass through to the AD Cmdlets
+
+.PARAMETER Objects
+    Specific type of AD objects to query.
+
+.EXAMPLE
+    Set-MtAdCache
+
+    This example sets the cache of AD queries.
+
+.LINK
+    https://maester.dev/docs/commands/Set-MtAdCache
+#>
+function Set-MtAdCache {
+    [CmdletBinding()]
+    [Diagnostics.CodeAnalysis.SuppressMessageAttribute('PSUseShouldProcessForStateChangingFunctions', '', Justification = 'Internal function')]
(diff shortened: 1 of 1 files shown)


powershell/public/Connect-Maester.ps1
```diff
@@ -116,11 +122,44 @@
 
    $__MtSession.Connections = $Service
 
-   $OrderedImport = Get-ModuleImportOrder -Name @('Az.Accounts', 'ExchangeOnlineManagement', 'Microsoft.Graph.Authentication', 'MicrosoftTeams')
+   $OrderedImport = Get-ModuleImportOrder -Name @('ActiveDirectory', 'Az.Accounts', 'ExchangeOnlineManagement', 'Microsoft.Graph.Authentication', 'MicrosoftTeams')
```
SamErde, line 125: The 'ActiveDirectory' module is not needed in this array. An inline explanatory comment would be helpful: it is checking modules that rely on the MSAL to determine which one should be connected to first, which ensures that the newest (and backwards compatible) version of Microsoft.Identity.Client.dll is loaded to avoid version conflicts between the MS modules.

soulemike: Since the switch case is dependent on the array returned, does that work as intended?

tdcthosc:
This is an awesome addition and I'd love to chip in; at least with our approach. We should certainly support cloud based automations, as I expect many have done exactly that given that Maesters original target was M365.

I have also been keen on getting started with a module for Palo Alto, as I have many years of experience with Palo Alto NGFW (and Palo Alto Panorama) and Palo Alto Cortex XDR, but I have been reluctant due to the connectivity question and how to approach this. Firewalls are (often) not exposed on management interfaces, so I also think local runners would be ideal in this scenario. The next question is how to relay that to the cloud automation. Happy to work with you guys on this 👍

PR #1381, opened 2025-12-15
Add Entitlement Management tests MT.1107-MT.1111

nicowyss:
## Summary
Adds 5 new Maester Core Tests for Entra ID Identity Governance Access Packages.

## Tests Added
- MT.1107 – Access packages and catalogs should not reference deleted groups
- MT.1108 – Access packages should not reference inactive or orphaned assignment policies
- MT.1109 – Access package approval workflows must have valid approvers
- MT.1110 – No catalog should contain resources without any associated access packages
- MT.1111 – Catalog resources must have valid roles (no stale app roles or SPNs)

## Notes
- Read-only tests
- Tested with Invoke-Maester
- No tenant modifications


SamErde:
@nicowyss, these are great checks! Do you have any thoughts on the suggestions from the Copilot review? I haven't had a chance to review much lately, but this might give you a few things to test. Thanks!

SamErde:
> @SamErde Thanks for the feedback, I'm really looking forward to sharing these tests with everyone. I would like to point out that there may be a conflict with the Maester test numbers, which is shown here as a conflict. Two test numbers: MT.1111
> 
> Although I claimed my numbers according to the procedure, see: “#️⃣ Pick next Maester test sequence number #697”

Ah, sorry to see that! Looks like innocent human error. You can take MT.1106 to replace MT.1111.

nicowyss:
> Awesome work @nicowyss
> 
> I believe these are the first ones around ID Governance in Maester. Thanks so much!!

Yes they are and hopefully this encourage more people to collaborate. Very welcome when are those url's reachable?
https://maester.dev/docs/tests/MT.1108

I would like to post something about thoose tests.

PR #1782, opened 2026-05-15
fix(MT.1066): ignore 'None' sentinel value for workload identity poli…

### LINKED ISSUE #1778: 🪲 MT.1066: False positive for Conditional Access policies targeting workload identities only
### Pre-submission checklist

### Describe the bug

Test MT.1066 ("Conditional Access policies should not reference non-existent users, groups, or roles") produces a false positive for Conditional Access policies that are scoped exclusively to workload identities. The test incorrectly flags such a policy as referencing a non-existent object of type "User" with the value "None", even though the policy intentionally contains no user, group, or role assignments.

The test result shows:

- Non-existent object: None
- Object type: User
- Condition: include

However, inspecting the raw policy data via browser developer tools confirms that the users object in the policy contains:

`"users": {
    "includeUsers": ["None"],
    "excludeUsers": [],
    "includeGroups": [],
    "excludeGroups": [],
    "includeRoles": [],
    "excludeRoles": [],
    "includeGuestsOrExternalUsers": null,
    "excludeGuestsOrExternalUsers": null
}`

The value "None" in includeUsers is the expected and valid API representation for a policy that targets workload identities and explicitly excludes all users. It is not a reference to a non-existent user object and should not be treated as one.

### Steps to reproduce

1. Create a Conditional Access policy in Entra ID scoped to "Workload identities" (e.g., "All owned service principals").
2. Ensure the policy has no users, groups, or directory roles assigned in the include/exclude conditions.
3. Run Maester and execute test MT.1066.
4. Observe that the policy is incorrectly flagged as referencing a non-existent user object with the value "None".

### Expected behavior

Test MT.1066 should recognize that "None" in the includeUsers array is a valid placeholder value used by the Microsoft Graph API to indicate that no users are targeted (as expected for workload identity-scoped policies). Such policies should not be flagged as referencing non-existent objects and should pass the test.

### Debug output

```text
Debug output not yet available. Will be added upon request if needed.
```

### Module version

```text
Maester Version: 2.1.0
```

### Environment data

```text
The test was executed inside a Docker container via GitLab CI/CD:

Docker Image: mcr.microsoft.com/microsoftgraph/powershell
Image Digest: sha256:6861ce525a317a6934b91771fa3df11b73e21409d32761d32362559d1c359e77
GitLab Runner Version: 18.11.1 (5265d41d)
Executor: Docker
```

### Screenshots

_No response_

### Additional context

- The policy is configured with "What does this policy apply to?" set to "Workload identities" and "All owned service principals" selected under Include.
- This appears to be a logic gap in the test where the literal string "None" returned by the Graph API for workload identity policies is being misinterpreted as a reference to a non-existent user object, rather than being recognized as a special sentinel value.

seick:
…cies

## 📑 Description
Fixes a false positive in test MT.1066 where Conditional Access policies 
scoped exclusively to workload identities were incorrectly flagged as 
referencing non-existent users.

When a policy targets workload identities only, the Microsoft Graph API 
returns `"None"` as a sentinel value in the `includeUsers` array. The test 
was attempting to look up a user with the ID `"None"`, which naturally 
returns a 404 and triggers a false positive.

The fix adds `"None"` to the list of known sentinel values that are already 
being filtered (alongside `"All"` and `"GuestsOrExternalUsers"`).

Closes #1778 

## ✅ Checks

## ℹ️ Additional Information

The `"None"` sentinel value is used by the Graph API specifically for 
policies targeting workload identities, where no users, groups, or roles 
are assigned by design.

Verified fix locally against a tenant with a workload identity-scoped 
Conditional Access policy, test now passes correctly.

Before fix:
mt1066_before-fix

After fix:
mt1066_after-fix

---

## How to Contribute

🏗️ Read our full [contributing guide](https://maester.dev/docs/contributing) for the Maester project.  
🧪 We also have additional instructions and a checklist for [creating tests](https://maester.dev/docs/contributing#checklist-for-writing-good-tests).  

Join us at the Maester repository [discussions](https://github.com/maester365/maester/discussions) or [Entra Discord](https://discord.maester.dev/) for more help and conversations!  
While you wait for a review, why not spread some Maester love on social media? Thank you! 💖

### DIFF
--- a/powershell/public/maester/entra/Test-MtCaReferencedObjectsExist.ps1
+++ b/powershell/public/maester/entra/Test-MtCaReferencedObjectsExist.ps1
@@ -33,7 +33,7 @@ function Test-MtCaReferencedObjectsExist {
         $policies = Get-MtConditionalAccessPolicy
 
         # Collect all referenced objects
-        $allUsers = $policies.conditions.users.includeUsers + $policies.conditions.users.excludeUsers | Where-Object { $_ -ne 'All' -and $_ -ne 'GuestsOrExternalUsers' -and $_ -ne $null -and $_ -ne '' } | Select-Object -Unique
+        $allUsers = $policies.conditions.users.includeUsers + $policies.conditions.users.excludeUsers | Where-Object { $_ -ne 'All' -and $_ -ne 'GuestsOrExternalUsers' -and $_ -ne 'None' -and $_ -ne $null -and $_ -ne '' } | Select-Object -Unique
         $allGroups = $policies.conditions.users.includeGroups + $policies.conditions.users.excludeGroups | Where-Object { $_ -ne $null -and $_ -ne '' } | Select-Object -Unique
         $allRoles = $policies.conditions.users.includeRoles + $policies.conditions.users.excludeRoles | Where-Object { $_ -ne $null -and $_ -ne '' } | Select-Object -Unique
 


Mynster9361 (approved):
LGTM

SamErde:
Could there be any CA policy scenarios that don't target users or workload identities at all? Is there a risk of those scenarios creating a similar kind of false positive (or false negative) when a policy is assessed--or is the lack of an explicitly specified `user or agents` precisely what shifts the scope to focus on workload identities?

seick:
> Could there be any CA policy scenarios that don't target users or workload identities at all? Is there a risk of those scenarios creating a similar kind of false positive (or false negative) when a policy is assessed--or is the lack of an explicitly specified `user or agents` precisely what shifts the scope to focus on workload identities?

Hi, here's my understanding, happy to be corrected if I'm missing an edge case:

According to the [Microsoft Graph documentation](https://learn.microsoft.com/en-us/powershell/module/microsoft.graph.identity.signins/new-mgidentityconditionalaccesspolicy), the includeUsers property accepts "User IDs in scope of policy unless explicitly excluded, None, All, or GuestsOrExternalUsers." So the only non-ID sentinel values are None, All, and GuestsOrExternalUsers. The existing filter already covers All and GuestsOrExternalUsers. None was the missing one. Since it's a documented sentinel rather than a user reference, filtering it shouldn't introduce false negatives.

On the broader scope question: a policy can also be scoped via includeGroups or includeRoles instead of includeUsers. In those cases includeUsers is just an empty array. But an empty array doesn't cause the "bug", because there's nothing for the loop to look up.

The "None" value seems to show up specifically when a policy targets workload identities via clientApplications. So I think the answer to your question is yes, when there's no user/group/role assignment, the scope shifts to clientApplications and the API returns "None" on the user side. I haven't seen a policy that targets neither, and I'd guess the API wouldn't allow it..

SamErde (approved):
OK, LGTM! Thanks for the research!

PR #1804, opened 2026-05-28
Allow configurable test settings

### LINKED ISSUE #1798: 🙏 Configurable accepted deviations for tests
### Describe the feature

I would like maester to be configurable, so it can ignore specific findings as accepted.
For example in my environment the test `CIS.M365.1.2.1` flagged multiple M365 groups, because they are public. The definition is "Ensure that only organizationally managed/approved public groups exist" - it doesn't state that public groups must not exist. It would be perfectly fine if an organization had public M365 groups, if they are aware of the risks and they control them. But right now the test is binary: If there are any public M365 groups, the test will fail.
I think it would make sense to allow the user to configure test-specific exceptions (allowed deviations) - similar to the configuration of `EmergencyAccessAccounts` accounts in `maester-config.json`.

For example we could extend the test related config with a `AcceptedEntities` property or something:
```json
    {
      "Id": "CIS.M365.1.2.1",
      "Severity": "Medium",
      "Title": "(L2) Ensure that only organizationally managed/approved public groups exist",
      "AcceptedEntities": [
        "0c3836ab-09e4-4a27-aa0a-da1933538b03",
        "27cfcc4e-c587-4da3-88dd-2997129ef12f"
      ]
    },
```

### Impact and importance

This would allow the users to continue to use pre-made tests without the need of rewriting the tests. If the architecture allows it, users can customize testing without changing PowerShell code.

### Alternatives you've considered

It would be possible to either not run a specific test, or to write an own alternative - but I don't think that's feasible for a lot of users.

### Additional context

_No response_

diecknet:
## 📑 Description

Previously `Get-MtMaesterConfig` only imported the severity from the test specific settings from `maester-config.json`. 
https://github.com/maester365/maester/blob/dc49768e3e9e645b94b0a6ee05d704cd8ef02eed/powershell/internal/Get-MtMaesterConfig.ps1#L153-L154

With this PR the cmdlet imports all values for a test from the JSON config file.

Closes #1798

## ✅ Checks

- [x] My code requires changes to the documentation.

## ℹ️ Additional Information

---


merill:
Thanks for this. 

So the apprach we want to take is to be able to define the configurable paramters in the Test itself as parameters. It's a little hard to do today with the Pester format. 

There is a bit of a rewrite we want to do and move to attribute based tests. Also the single file maester-config, feels like a bad idea given our experience having to keep merging it all the time with merge conflicts.

merill:
I think we should close off this PR and take the design back to the drawing board to come up with a better way.

diecknet:
Thanks for taking a look at this PR @merill :)
Since this will probably not get merged then - good thing I didn't find the time to finish the documentation stuff and adding a configuration for a test (as kind of proof of concept) 😅
Anyway - I'm looking forward to another approach. Not sure if I can help with that, because I'm not really well versed with Pester.

merill:
Super sorry about this and my bad for not getting back to you early.

PR #1810, opened 2026-05-30
Remove validation of parameter $Server

JeanPhilippeGeorge:
## 📑 Description
Fix error in 'CISA.MS.EXO.2.2: An SPF policy SHALL be published for each domain, designating only these addresses as approved senders.'

function Resolve-SPFRecord

Comment [ValidateNotNullOrEmpty()] on the parameter $Server and add a comment explaining why. That parameter is not mandatory, it should not be validated for not null or empty as it can be null or empty.

If anybody has a better text for the comment, I'm happy to change it.

## ✅ Checks
   1 test fails but there is no indication of which one ... and it seems to succeed on the github side ...    
    Checked as the build pass in the CI

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated mail authentication DNS/SPF resolution to use the custom DNS server only when an explicit DNS server IP value is provided, preserving default DNS behavior otherwise
  * Added automated test coverage to verify `Resolve-SPFRecord` is called with no `Server` parameter when DNS server is omitted or null, and with the expected `Server` value when provided


SamErde:
Does the `ValidateNotNullOrEmpty()` method truly force a **Server** to be specified every time? @claude, review this suggested change. (@merill, have we added the Claude app to the repository so it can perform reviews? If we do, would it use your tokens or tokens from the person requesting the review?) 🙂

SamErde:
@JeanPhilippeGeorge, thanks for your suggestion. In this case, using the `ValidateNotNullOrEmpty()` method does not cause a problem. The combination of `Mandatory = $false` and `[ValidateNotNullOrEmpty()]` is essentially saying: "this parameter is optional, but if you do provide it, it must have a real value." Therefore, it is a relevant validation step that we believe can and should stay unless you have found an edge case that somehow triggers it. 🙂

JeanPhilippeGeorge:
> @JeanPhilippeGeorge, thanks for your suggestion. In this case, using the `ValidateNotNullOrEmpty()` method does not cause a problem. The combination of `Mandatory = $false` and `[ValidateNotNullOrEmpty()]` is essentially saying: "this parameter is optional, but if you do provide it, it must have a real value." Therefore, it is a relevant validation step that we believe can and should stay unless you have found an edge case that somehow triggers it. 🙂

Completely agree. But ... see next comment

JeanPhilippeGeorge:
@SamErde please reopen the PR.

I don't think I'm in an edge case. This test is always erroring for me (not failing). On various pcs. And looking at the code, I can't see a way that is can work.

Does that code work for you? Either passing or failing depending on spf state.


With ValidateNotNullOrEmpty, the test error:
image


Without ValidateNotNullOrEmpty, it succeeds:
image



For using or not ValidateNotNullOrEmpty I think it's a convenience:
+ Either you use it and every caller of the function has to check if the $Dns Server is null or not before passing it
+ Or you don't and the final code check if the parameter id defined or not. Which Resolve-SPFRecord is already doing.

Personaly I prefer to test once at the end, but as there is only one caller (Get-MailAuthenticationRecord), I'm happy to test in Get-MailAuthenticationRecord

And Get-MailAuthenticationRecord accept a $DnsServerIpAddress and don't validate it. So for consistency it seems judicious to add ValidateNotNullOrEmpty to the parameter $DnsServerIpAddress

To recap:
```
# CISA.MS.EXO.2.2: An SPF policy SHALL be published for each domain, designating only these addresses as approved senders. 
# Test-MtCisaSpfDirective.Tests.ps1
# calls Test-MtCisaSpfDirective
  $cisaSpfDirective = Test-MtCisaSpfDirective # No DnsServer

# Test-MtCisaSpfDirective.ps1
  $spfRecord = Get-MailAuthenticationRecord -DomainName $domain.DomainName -Records SPF # No $DnsServer

# Get-MailAuthenticationRecord.ps1
# Has this parameter (not mandatory, no ValidateNotNullOrEmpty)
  # DKIM DNS record Name to retrieve.
  [string]$DkimDnsName,

# Then this call does not test if $DnsServerIpAddress is defined or not. And as Resolve-SPFRecord want either no parameter or a defined one it fails
  $recordSet.spfLookups = Resolve-SPFRecord -Name $DomainName -Server $DnsServerIpAddress
```

I will make a new commit, removing my change in Resolve-SPFRecord.ps1 and changing the previous line to this
```
  if ($DnsServerIpAddress) {
    $recordSet.spfLookups = Resolve-SPFRecord -Name $DomainName -Server $DnsServerIpAddress
  } else {
    $recordSet.spfLookups = Resolve-SPFRecord -Name $DomainName
  }
```

Tell me what you think.

PS: I cannot commit when the PR is closed ...

JeanPhilippeGeorge:
I added the Copilot suggestion

powershell/public/cisa/exchange/Get-MailAuthenticationRecord.ps1
```diff
@@ -73,7 +74,6 @@
 
         $splat = @{
             DomainName         = $DomainName
-            DnsServerIpAddress = $DnsServerIpAddress
```
SamErde, line 76: A common approach to using clean, dynamic splats looks like this:

```powershell
# Create the "always present" parameters.
$LookupSplat = @{
    DomainName   = $DomainName
    QuickTimeout = $QuickTimeOut
    NoHostsFile  = $NoHostsFile
}
# Add the `DnsServerIpAddress` parameter IF it has been specified by the user.
if ($PSBoundParameters.ContainsKey('DnsServerIpAddress')) {
    $LookupSplat.DnsServerIpAddress = $DnsServerIpAddress
}
```
Then, you can simplify the next `Resolve-SPFRecord` line to only contain:
```powershell
$RecordSet.SpfLookups = Resolve-SPFRecord @LookupSplat
```

In this case, the parameter definition already validates that a null or empty `DnsServerIpAddress` has _not_ been allowed. Then, the `$PSBoundParameters.ContainsKey()` detects the user's/caller's intent: was a value explicitly provided?

JeanPhilippeGeorge: Resolve-SPFRecord don't have the same parameters as $LookupSplat (2 with different names, QuickTimeOut does not exists), so no easy splatting here.

I will change the test for the parameter DnsServerIpAddress as you suggested.

To be honest, I know that it is the more explicit / cleaner way to test if a parameter is provided, but I never remember that it exists before seeing it ;-) Maybe this time I will remember!

JeanPhilippeGeorge:
Possible duplicate of https://github.com/maester365/maester/pull/1845

merill (approved):
LGTM

PR #1814, opened 2026-06-01
Refine compliance check for managed domains

brianreidc7:
Filter out managed domains that support SharePoint (M365 old public hosted domains product) and adjust compliance check for password validity period as this is a nullable integer, so where domain was previously federated and no longer federated, the value can be null but this means password does not expire. Update the markdown reporting for skipping the SharePoint domains

## 📑 Description

Closes # 

## ✅ Checks

- [ ] My pull request adheres to the code style of this project.

## ℹ️ Additional Information

---

## How to Contribute

🏗️ Read our full [contributing guide](https://maester.dev/docs/contributing) for the Maester project.  
🧪 We also have additional instructions and a checklist for [creating tests](https://maester.dev/docs/contributing#checklist-for-writing-good-tests).  

Join us at the Maester repository [discussions](https://github.com/maester365/maester/discussions) or [Entra Discord](https://discord.maester.dev/) for more help and conversations!  
While you wait for a review, why not spread some Maester love on social media? Thank you! 💖

## Release Notes

* **Bug Fixes**
  * Updated password expiration compliance logic to exclude legacy managed domains that only support SharePoint from the compliance requirement.
  * Tightened validation so password expiration is considered compliant only when explicitly set to “never expire” (the maximum sentinel value).
  * Improved compliance reporting: enhanced the markdown output with a new reason field and clearer per-domain pass/skip/fail explanations.
* **Tests**
  * Added Pester coverage for the updated compliance and reporting behaviors, including legacy and edge-case domain handling.


merill (changes requested):
@brianreidc7 Tx a lot for the PR.

If these are SharePoint only domains then should we use `-eq` in line 42 `supportedServices -notcontains "SharePoint"`?

What do you see in your tenant's data?

This also lines with the -eq you introduced in line 78.

Given we are adding a new type here. It might be good to include a new Reason column after Validation and include the reason we are skipping (speficially your call out about this being for legacy SharePoint domains) 

-----

On the password expiry, are you sure when it is null that it is the same as 'Never Expire'? In the docs it says `If the property isn't set, a default value of 90 days is used.` See https://learn.microsoft.com/en-us/graph/api/resources/domain?view=graph-rest-1.0#properties

Even if it were not the default, I belive we should err on the side of the setting this value explicitly.

brianreidc7 (commented):
Updated to support legacy domains and added a new output column for this

merill (approved):
LGTM

Issue #1748, opened 2026-05-07
Defender Results dont consider multi Policy Tenancies.


Gbenrotto:
The Defender results dont consider Scenarios where something is properly configured on a separate policy. It should evaluate all policies and then plan based on that.

merill:
Closing this as a duplicate of #1922

PR #1841, opened 2026-06-13
fix(orca): correct BuiltIn property name in ORCA113 and add missing guard in ORCA179

### LINKED ISSUE #1806: 🪲 ORCA.179: Safe Links is enabled intra-organization Fail even though enabled in custom policy
### Pre-submission checklist

### Describe the bug

When I run "ORCA.179: Safe Links is enabled intra-organization" if Fail even though I created a custom policy (below called "Default Safe Links")

Image

It says in the description "The built-in policy is ignored in this check" but that is not true. This fails on all my tenants but there is one tenant (a test tenant) where it Pass and that is a tenant where the Preset Security Policies Standard/Strict policies are enabled:

Image

If I disable the "Preset Security Policies" it Fails again. Running Maester 2.1.0.

### Steps to reproduce

1. Create a custom policy with Safe Links enabled
2. Make sure Preset Security Policies Standard/Strict are NOT enabled
3. Run Maester
4. Check if "ORCA.179: Safe Links is enabled intra-organization" fail on the "Built-In Protection Policy"

### Expected behavior

It should "Skip" the "Built-In Protection Policy"

### Debug output

```text

```

### Module version

```text
2.1.0
```

### Environment data

```text
Version
7.4.16

Edition
Core

Platform
Unix (running i Azure DevOps)
```

### Screenshots

_No response_

### Additional context

Seems to be the same behaviour on the following test but that is for another bug :)
ORCA.113: AllowClickThrough is disabled in Safe Links policies.

Jhope188:
## Problem

`ORCA.113` and `ORCA.179` produce false failures on correctly configured tenants because they evaluate the **Built-In Protection Policy**, which is Microsoft-controlled and cannot be modified by tenant admins.

**ORCA.113 — wrong property name (`IsBuiltIn` vs `BuiltIn`)**

The check already had the intent to exclude the Built-In Policy:
```powershell
if(!$Config["PolicyStates"][$Policy.Guid.ToString()].IsBuiltIn)
```
However, `IsBuiltIn` does not exist on the `PolicyInfo` class — the correct property is `BuiltIn` (see `orcaClass.psm1` line 288). In PowerShell, accessing a non-existent property returns `$null`, and `!$null` evaluates to `$true`, so the condition always passes and the Built-In policy is **always** included in the evaluation. The intended exclusion was silently broken.

**ORCA.179 — no guard at all despite documented intent**

The `Importance` description explicitly states:
> *"The built-in policy is ignored in this check, as it only provides the minimum level of protection."*

But the `GetResults` method has no code to skip the Built-In policy. Every Safe Links policy including the read-only Built-In Protection Policy is evaluated. Since the Built-In policy has `EnableForInternalSenders=False` and that setting cannot be changed, any tenant without Standard/Strict Preset Security Policies will permanently fail this check regardless of their custom policy configuration.

## Root Cause

The Built-In Protection Policy is only suppressed in `Get-PolicyStateInt.ps1` when a Preset Security Policy (Standard/Strict) is applied (`TypeHasAppliedPresetPolicy = $true`). On tenants using only custom policies, the Built-In policy remains in the evaluated set with `Applies=True` and `BuiltIn=True`.

Individual checks are responsible for filtering it out when they intend to. ORCA.113 tried to but used the wrong property name. ORCA.179 did not filter it at all.

## Fix

**`check-ORCA113.ps1`** — one character change, `IsBuiltIn` → `BuiltIn`:
```powershell
# Before (broken — IsBuiltIn always returns $null → filter never fires)
if(!$Config["PolicyStates"][$Policy.Guid.ToString()].IsBuiltIn)

# After (correct property name)
if(!$Config["PolicyStates"][$Policy.Guid.ToString()].BuiltIn)
```

**`check-ORCA179.ps1`** — add the missing `Continue` guard at the top of the `ForEach` loop, consistent with the documented behaviour and with ORCA.113's intent:
```powershell
if($Config["PolicyStates"][$Policy.Guid.ToString()].BuiltIn)
{
    Continue
}
```

## Testing

Verified against tenant ConditionalAccess.tech with:
- Custom Safe Links policy: `AllowClickThrough=False`, `EnableForInternalSenders=True` ✅
- Built-In Protection Policy: `AllowClickThrough=True`, `EnableForInternalSenders=False` (read-only)
- No Standard/Strict Preset Security Policies enabled

Before fix: ORCA.113 and ORCA.179 fail (Built-In policy evaluated)
After fix: Only the custom admin-configurable policy is evaluated → both pass correctly

No regression on tenants without custom policies — the Default policy fallback path is unchanged, and the `PolicyCount -eq 0` guard still fires when there are genuinely no non-built-in policies configured.

## References

Closes #1806
Related to #1748

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated built-in policy identification in compliance checks to use the correct property reference.
  * Improved built-in policy filtering to ensure they are properly excluded from security evaluations.


merill (approved):
Thanks for the fix @Jhope188 

It's good that we are keeping the ORCA tests updated here.

PR #1842, opened 2026-06-13
fix(mt1020): null-safe role lookup and ABA service-principal detection

Jhope188:
## Problem

MT.1020 errors on some tenants with:

```
Get-MtRoleInfo: You cannot call a method on a null-valued expression.
```

It also silently misclassifies tenants using **Application-Based Authentication (ABA)** — introduced in Microsoft Entra Connect **v2.5.76.0** — where directory sync is performed by a service principal rather than a user account.

---

## Root Cause 1 — `$script:MtRoles` null on module reload

`Get-MtRoleInfo` calls `$script:MtRoles.ContainsKey()`. When the Maester module is reloaded inside an **existing PowerShell session**, the `MtRoleDefinition` class cannot be redefined (PowerShell restriction). The dot-source of `Get-MtRoleInfo.ps1` throws, the `psm1` `try/catch` swallows the error, and `$script:MtRoles` is never initialised. The function body from the *previous* load is still callable, but `$script:MtRoles` is `$null`, so `.ContainsKey()` throws.

`Test-MtCaExclusionForDirectorySyncAccount` then passes the `$null` return value directly into the mandatory `[guid[]]$RoleId` parameter of `Get-MtRoleMember`, causing the binding error and an `Error` result instead of a graceful skip.

## Root Cause 2 — ABA tenants not handled

As of Entra Connect **v2.5.76.0**, directory synchronisation can run under a registered **service principal** (ABA) rather than a dedicated sync user account. Service principals are not subject to Conditional Access policies, so there is nothing to exclude. Previously the function would fall through to the per-policy loop and either:
- Reach the `elseif ($memberIds.Count -eq 0)` branch silently with no descriptive message, or
- Fail to evaluate any policies because there are no user member IDs, yet still report pass/fail ambiguously.

---

### `Get-MtRoleInfo.ps1`
Add a `$null` guard before `.ContainsKey()`:
```powershell
if ($null -eq $script:MtRoles) { return $null }
```

### `Test-MtCaExclusionForDirectorySyncAccount.ps1`

1. **Null-safe role ID checks** — skip `Get-MtRoleMember` when `Get-MtRoleInfo` returns `$null`.
2. **Early ABA detection** — after resolving members, classify them as user vs. service-principal. If *all* sync members are service principals (ABA), return `$true` immediately with an informative message explaining the Entra Connect v2.5.76.0 ABA context.
3. **Simplification** — the per-policy `$memberIds` computation now reuses the pre-computed `$userSyncMembers` list instead of re-filtering `$Members` on each iteration.

---

## Testing

All 18 unit tests pass (2 new tests added for the null-role and ABA scenarios):

```
Tests Passed: 18, Failed: 0, Skipped: 0
```

Verified locally against a tenant running Entra Connect v2.5.76.0+ (ABA mode): MT.1020 now returns `Pass` with an informative message instead of `Error`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved role lookup resilience during module reloads.
  * Enhanced null-value handling in conditional access exclusion validation to prevent binding errors.

* **Tests**
  * Expanded test coverage for directory sync account exclusion scenarios, including service principal patterns.
  * Added test cases for role lookup failures.


merill (approved):
This looks good. Thakns @Jhope188 !!

PR #1844, opened 2026-06-15
fix: CIS.M365.4.1 skip when tenant is not licensed for Intune

Korthal-Maiyn:
## 📑 Description

`Test-MtCisDevicesWithoutCompliancePolicyMarked` (CIS.M365.4.1) calls
`Invoke-MtGraphRequest -RelativeUri "deviceManagement/settings"`. On a tenant
without an active Intune licence this Graph endpoint returns 400 Bad Request,
which is uncaught, so the test reports as errored instead of skipped.

Microsoft documents that the Intune Graph API requires an active Intune licence
for the tenant: https://learn.microsoft.com/graph/api/resources/intune-deviceconfig-devicemanagementsettings

Fix: add a licence guard before the Graph call, mirroring the existing pattern in
`Test-MtDeviceComplianceSettings`, which checks the same `secureByDefault` setting
through the same endpoint:

```powershell
if (-not (Get-MtLicenseInformation -Product Intune)) {
    Add-MtTestResultDetail -SkippedBecause NotLicensedIntune
    return $null
}
```

The test now reports Skipped (Not Licensed: Intune) on tenants without Intune
instead of erroring.

## ✅ Checks

## Summary by CodeRabbit

* **Bug Fixes**
  * Test now includes an upfront licensing check and gracefully skips if Intune is not licensed, preventing unnecessary execution.


merill (approved):
Good catch. Thanks @Korthal-Maiyn!

PR #1900, opened 2026-07-02
chore: add .gitattributes to normalize line endings

SamErde:
## TL;DR

Adds a repository-level `.gitattributes` so text files are stored with LF, preventing CRLF files from being committed from Windows and tripping linters (yamllint flagged this twice during review of #1899).

## Details

- `* text=auto` stores LF in the repository for all detected text files.
- YAML, shell, JS/TS, JSON, and markdown files are forced to LF in the working tree too, since CI tooling (yamllint, actionlint, prettier) expects Unix endings.
- PowerShell files (`.ps1`, `.psm1`, `.psd1`, `.ps1xml`) are stored with LF but keep platform-default working-tree endings so Windows contributors retain a native editing experience.
- Common binary asset types are marked `binary`.

**Scope note:** this PR intentionally does not renormalize existing files (no `git add --renormalize`), keeping the diff to a single new file. Existing CRLF blobs will be corrected naturally as files are next modified. A follow-up bulk renormalization can be done later if desired.

Follow-up to review feedback on #1899.

## Summary by CodeRabbit

* **Chores**
  * Added repository-wide line-ending rules to keep text files consistent across platforms.
  * Standardized common source and documentation formats to use LF endings.
  * Marked common binary asset types as non-text, while treating SVG files as text for improved handling.

PR #1901, opened 2026-07-02
ci: disable credential persistence in build-validation checkout

SamErde:
## TL;DR

Sets `persist-credentials: false` on the `actions/checkout` step in `build-validation.yaml`. This job only reads the repository and never pushes or fetches with the token after checkout, so persisting the credential in the git config is unnecessary. Low-cost hardening default (zizmor `artipacked` finding).

## Scope

Intentionally limited to `build-validation.yaml`. The publish workflows are excluded because they use `stefanzweifel/git-auto-commit-action` to commit website version bumps, which requires the persisted checkout credential.

Follow-up to review feedback on #1899. Minimal 4-line diff (the file's existing mixed line endings are left untouched here; normalization is proposed separately in #1900).

Suggested by CodeRabbit.

## Summary by CodeRabbit

* **Chores**
  * Improved the build validation process to prevent checkout credentials from being retained after repository access, reducing the risk of unintended token persistence.


merill (approved):
LGTM

Issue #1559, opened 2026-04-01
New Build Workflow and Performance Optimizations


SamErde:
As Maester continues to grow with more tests and features, its already long import time will also continue to increase. 

These updates will follow several steps to reduce the time required when importing the module and when discovering tests.

The following is an initial pass at a plan that will be used.

---

# Maester Module Optimization Plan

This document describes a set of targeted build-time and structural optimizations for the
`maester365/maester` repository. The goal is to reduce `Import-Module` time, preserve safe test packaging, and investigate test discovery performance separately ΓÇö without changing any user-facing
behavior or removing bundled tests.

All changes are **build-time transformations only**. The source tree stays as individual files
for developer ergonomics. The published artifact changes.

---

## Guiding Principles

- Source files are never modified directly by the build optimizations.
- The published module must be functionally identical to the current one.
- `Custom/` test folder is never touched by any build step.
- Load order (internal before public) must be explicitly enforced, not assumed.
- Base classes must always be defined before derived classes in consolidated output.

---

## Coding Standards

All PowerShell files created or modified as part of this project ΓÇö including the build script,
any updated module functions, and any new helper scripts ΓÇö must conform to the following
standards. These apply to new code and to any code that is substantively modified. They are
not retroactively enforced on untouched files in a single pass, but the style pass in Item 1
is an appropriate opportunity to correct obvious violations encountered along the way.

| Standard | Rule |
|---|---|
| Brace style | One True Brace Style (OTBS). Opening brace on the same line as the statement. |
| Casing | Pascal Case for all variable names, parameter names, and object properties. |
| Indentation | 4 spaces. No tabs. |
| Encoding | UTF-8 with BOM (`utf8BOM`) for all `.ps1` and `.psm1` files. |
| Function verbs | Approved PowerShell verbs only (`Get-Verb` is the authoritative list). |
| Trailing whitespace | No trailing spaces on any line. |
| Final newline | One blank line at the end of every file. |

**OTBS example:**
```powershell
function Invoke-MaesterBuild {
    [CmdletBinding()]
    param (
        [string] $OutputPath
    )

    if (-not (Test-Path -Path $OutputPath)) {
        New-Item -ItemType Directory -Path $OutputPath | Out-Null
    }
}
```

---

## 1. Move Comment-Based Help Inside Function Bodies

**Status:** Required prerequisite for Item 2. Must be completed first.

**Problem:** Many `.ps1` files in `powershell/public/` and `powershell/internal/` have
[...]

PR #1930, opened 2026-07-13
fix: restore test result markdown in consolidated module

### LINKED ISSUE #1924: Reports include PowerShell function definitions
## Description

JSON and Excel reports can contain PowerShell function definitions in the test result details instead of only the expected result text.

Discord report: https://discord.com/channels/1524331650337603584/1524341702981193750/1524756850992615516

## Reported environment

- Maester 2.1.70 (the reporter also suspected 2.1.52)
- PowerShell 7.6.3
- Pester 6
- Rolling back to Maester 2.1.0 removes the problem

## Root cause

The regression starts with Maester 2.1.66, when the release workflow began publishing the consolidated `Maester.psm1` artifact. Consolidation removes the script-scope boundaries provided by individual function files. A conditionally initialized `$result` can then resolve to Pester/mock state, causing the `Add-MtTestResultDetail` function definition to be inserted into `%TestResult%`.

## Expected behavior

Report result details contain only the test output and never PowerShell function source.

## Proposed fix

Publish a source-layout package that preserves PowerShell script-scope boundaries, while retaining the consolidated artifact for optimization validation.

Tracked by PR #1923.

merill:
## Summary

Fixes reports that displayed PowerShell source instead of test details and restores companion Markdown descriptions in the consolidated published module, including tests excluded by tag filters.

## Root cause

The published module consolidates individual function files into `Maester.psm1` but did not include their companion `.md` files. As a result, `$MyInvocation.PSCommandPath` no longer identified the original function file. The previous string replacement left the `.psm1` path unchanged, causing the entire module source to be read as Markdown.

Filtered `NotRun` tests never execute `Add-MtTestResultDetail`. Report conversion therefore had no result details and deliberately fell back to rendering the raw Pester script block.

A small number of empty-data paths also substituted an uninitialized `$result`, allowing a parent-scope value to leak into `%TestResult%`.

## Changes

- Generate `Maester.TestMetadata.json` during the module build from all paired `.ps1` and `.md` files.
- Load and cache the metadata bundle on the first lookup in the published module.
- Use one internal resolver for generated JSON and adjacent source Markdown.
- Resolve metadata for filtered tests from commands in their discovered Pester script block.
- Resolve EIDSCA dispatcher tests from their control ID.
- Give `NotRun` tests a neutral detail instead of raw PowerShell when no companion Markdown exists.
- Preserve adjacent `.md` fallback for source development without a generated bundle.
- Add `Build-LocalMaester.ps1` so developers can build, validate, and globally import the local artifact with one command.
- Configure VS Code F5 to run the same build-and-load workflow.
- Document the consolidated build workflow and the separate source-level breakpoint workflow in the contributor guides.
- Validate that the published metadata bundle exists, parses, and contains entries.
- Initialize `$result` in `Test-MtCisaDlp`, `Test-MtCisaDlpPii`, and `Test-MtCisaExternalSenderWarning`.

## Validation

- Ran `Build-LocalMaester.ps1` end to end and confirmed `Invoke-Maester` remains loaded from `./module` after the helper exits.
- Validated the VS Code launch configuration and its exact PowerShell command.
- Compiled the documentation successfully; full static generation reached an unrelated existing `useDocsSidebar` failure on the command index pages.
- Built and validated the consolidated module successfully.
- Confirmed the generated bundle contains all 388 companion Markdown entries.
- Reproduced `Invoke-Maester -Tag MS.SHAREPOINT` against both the source and built modules.
- Confirmed `AZDO.1000` and `EIDSCA.ST08` show their authored Markdown while remaining `NotRun`.
- Confirmed every `NotRun` entry has safe details and none fall back to raw source.
- Confirmed the built module imports and exports 419 commands.
- Ran all 6,096 PowerShell tests with zero failures.

Closes #1924.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Description/Result templating for test results, including filtered/NotRun scenarios, and ensured deterministic `%TestResult%` replacement across generated reports.
* **New Features**
  * Module builds now generate `Maester.TestMetadata.json` from Markdown fixtures; added `Build-LocalMaester.ps1` for local build/validation/load.
* **Build & Release**
  * Validation now fails fast when `Maester.TestMetadata.json` is missing or empty, with updated check ordering.
* **Tests**
  * Updated fixtures and assertions to cover the new metadata artifact and rendering behavior.
* **Documentation**
  * Refreshed README, contributing workflow, and VS Code build/load configuration.


merill:
Addressed the current automated review feedback in `4580de63`:

- Replaced the three new build-script `Write-Host` calls flagged by GitHub Advanced Security with suppressible `Write-Information` output.
- Switched `%TestResult%` substitution to literal `.Replace()` so values such as `$1,000` remain intact.
- Made synthetic `NotRun` result details use the full eight-property schema.
- Tightened the contributor-guide wording.

Older XSPM and anti-spam findings refer to pre-existing code that is no longer part of this minimized PR diff, so they were intentionally not folded into #1930. The regression-test thread is obsolete because that earlier test file was removed during scope reduction.

Validation: consolidated build/import passed, all 406 filtered `NotRun` details had the complete schema, literal dollar-value substitution passed, and all 6,096 PowerShell tests passed. All inline review threads are resolved.

SamErde (commented):
Thanks for turning this around so quickly — I dug into the mechanics of this fix against #1924 and the consolidation plan (#1559/#1899), and I think the metadata-bundle design is the right call versus reverting consolidation. The `.psm1` extension guard in the resolver also structurally prevents the original self-read failure from recurring. I verified the core behaviors with in-memory probes on both PowerShell 7.6 and Windows PowerShell 5.1 (call-stack frame semantics, `ConvertFrom-Json` on the bundle size, case-insensitive property lookup, the `Get-Variable -Scope Script` negative cache) and they all check out.

One gap I'd like to see closed before merge, one small suggestion in the same function, and a couple of FYIs.

## 1. Custom-test companion markdown no longer resolves (request: fix in this PR)

The documented custom-test pattern in [Writing tests → Advanced guide → "Splitting tests into multiple files"](https://maester.dev/docs/writing-tests/advanced-concepts#splitting-tests-into-multiple-files) is:

```powershell
# Custom/ContosoUsers.Tests.ps1
BeforeAll {
    . $PSScriptRoot/Test-ContosoUsersMissingManagers.ps1   # defines the function in Pester's container scope
}
```

with `Custom/Test-ContosoUsersMissingManagers.ps1` calling `Add-MtTestResultDetail -Result ...` (docs explicitly say **not** to pass `-Description`) and `Custom/Test-ContosoUsersMissingManagers.md` alongside it.

That pattern still works on the affected releases (2.1.66–2.1.70), because for custom tests the caller's file is the user's on-disk `.ps1`, so the old `$MyInvocation.PSCommandPath` lookup finds the adjacent `.md`. On this branch it breaks:

- The function isn't in `Maester.TestMetadata.json` (it's a user file, not built into the bundle), and
- the fallback `Get-Command -Name $name -CommandType Function` runs **inside the Maester module**, and command discovery there only sees module + global commands. A function dot-sourced in `BeforeAll` lives in Pester's container scope, not global — so `Get-Command` returns nothing. (Probed: a caller-scope function is invisible; only `global:` functions resolve.)

Net effect: `$metadata` is `$null` and the authored markdown is silently dropped for every custom test written per the docs — a new regression relative to today's broken builds.

The good news: the calling frame already carries the file we used to key on. `$callStack[1].ScriptName` returns exactly what `$MyInvocation.PSCommandPath` returned for file-defined callers (probed, including the dot-sourced-into-a-child-scope case), and for module-internal callers in the consolidated module it returns `...\Maester.psm1`, which the existing `.ps1` extension guard rejects — so #1924 can't come back through this path. Sketch:

```powershell
# Add-MtTestResultDetail.ps1
$callStack = @(Get-PSCallStack)
$callerName = if ($callStack.Count -gt 1) { $callStack[1].Command } else { $null }
$callerFile = if ($callStack.Count -gt 1) { $callStack[1].ScriptName } else { $null }
$metadata = Get-MtTestResultTemplate -CommandName $callerName -SourceFile $callerFile
```

```powershell
# Get-MtTestResultTemplate.ps1
param(
    [Parameter()] [string[]] $CommandName,
    [Parameter()] [string] $TestId,
    [Parameter()] [string] $SourceFile
)
...
# (the early "return $null when no candidate names" needs to also consider $SourceFile)

# After the bundle + Get-Command lookups miss:
if ($SourceFile -and [System.IO.Path]::GetExtension($SourceFile) -eq '.ps1') {
    $markdownPath = [System.IO.Path]::ChangeExtension($SourceFile, '.md')
    if (Test-Path -LiteralPath $markdownPath) {
        # same read/split as the Get-Command fallback
    }
}
```

This also restores the other historical variant (a `.md` next to a `.Tests.ps1` that calls `Add-MtTestResultDetail` directly from an `It` block — there the frame's `Command` is `<ScriptBlock>` but `ScriptName` is the `.Tests.ps1` path). A unit test that dot-sources a function into a child scope with an adjacent `.md` (TestDrive works) would lock the custom-test contract in.

## 2. Suggestion: literal `Replace` in the `$hasGraphResults` path too

This PR correctly switched the markdown-template substitution to literal `String.Replace`, but the graph-objects block a few lines below still uses regex `-replace "%TestResult%", $graphResultMarkdown`. The replacement text there is built from tenant data (policy/user display names), and .NET substitution tokens in it corrupt the output — probed: a name containing `$_` injects the entire input string; `` $` ``, `$'`, `$&`, `$$` also mangle (`$1` happens to be safe since there are no capture groups). Since this function is already being edited here, converting that one line too would close out the same bug class in this function. The remaining ~60 `-replace "%TestResult%"` sites across `public/` could be a mechanical follow-up sweep.

## 3. Minor robustness (non-blocking)

- The old markdown block wrapped discovery in `try/catch` and degraded to a `Write-Warning`; the new `Get-PSCallStack` + `Get-MtTestResultTemplate` calls sit outside any `try/catch`, so an unexpected resolver failure now fails the test itself rather than just losing the description. Cheap to restore parity.
- In `ConvertTo-MtMaesterResult`, the NotRun branch dereferences `$test.ScriptBlock.Ast` without a null guard — if Pester ever hands back a test without a script block, report conversion throws. A `$test.ScriptBlock -and ...` guard keeps it safe.

## 4. Related findings outside this PR's scope (FYI)

While auditing the consolidation for other layout-dependent runtime behavior, I found one more latent issue in the shipped builds, unrelated to this PR: the build script's hardcoded `$__MtSession` header (`build/Build-MaesterModule.ps1`, Phase C preamble) is missing the `SpoCache` and `GitHubCache` keys that `powershell/Maester.psm1` initializes. In 2.1.66+ a direct `Get-MtSpo` call throws `Cannot index into a null array` (the read at `Get-MtSpo.ps1:35`, both PS editions), and `Invoke-MtGitHubRequest` would fail the same way before `Connect-MtGitHub` runs. `Invoke-Maester` runs are unaffected because `Clear-ModuleVariable` re-creates both keys. I'd suggest the build extract the `$__MtSession` block from the source `Maester.psm1` instead of duplicating it, plus a key-parity check in `Test-MaesterModuleOutput.ps1`. Happy to file this separately with details.

Also noted for a future hygiene pass (not blockers here): `Test-MtCisaSpoSharingAllowedDomain.ps1:39` and `Test-MtCisaAntiSpamAllowList.ps1:41` assign `$result =` inside their loops, so only the last domain/policy renders in the detail table; `Test-MtCisaAntiSpamSafeList.ps1:42` does a `-replace "%TestResult%"` but neither markdown branch contains the placeholder (dead line, and `$result` is never assigned in that function); `Test-MtCisaSpoSharing.ps1` would benefit from the same `$result = ''` init this PR added to the three CISA exchange tests.

---

With item 1 addressed (and the branch updated — it's currently behind main), this looks good to me. The validation you ran on both source and built layouts plus the new bundle checks in `Test-MaesterModuleOutput.ps1` give good confidence for the shipped suites.

SamErde:
One additional guardrail would make this fix durable: please add an end-to-end regression test against the final consolidated artifact, rather than only validating the metadata bundle's shape.

`Test-MaesterModuleOutput.ps1` currently verifies that `Maester.TestMetadata.json` exists, parses, and is non-empty, but it never imports the built module and exercises `Add-MtTestResultDetail`. That is the behavioral gap that allowed #1924 to pass build validation.

Could the test build and import `./module/Maester.psd1` in a clean process, invoke one stable markdown-backed helper through a deterministic skipped or mocked path, and assert that its rendered description matches the expected companion markdown and does not contain raw module or function source? This would make the published artifact's runtime behavior part of the build contract and directly prevent this regression from recurring.

merill:
Tx @SamErde that was a good catch. 

I've also added the regression tests and addressed all three...

- Restored adjacent companion Markdown for documented custom tests by passing the caller's `ScriptName` into the shared metadata resolver.
- Switched Graph `%TestResult%` substitution to literal `String.Replace` so replacement tokens in tenant data remain intact.
- Added a runtime metadata probe to `Test-MaesterModuleOutput.ps1`; validation now imports the consolidated module, invokes `Add-MtTestResultDetail`, and compares the rendered result with the generated bundle.
- Added artifact-level Pester regressions that build and import the consolidated module, invoke `Test-MtCisaDlp` through deterministic mocks, assert the authored Markdown is rendered, and assert raw function source is absent.
- Added regressions for adjacent custom-test Markdown and literal Graph replacement tokens.

Validation completed:

- Focused build/result suites: 7 passed.
- PSScriptAnalyzer plus published-module regressions: 685 passed.
- Full consolidated build and `Test-MaesterModuleOutput.ps1` validation passed with 398 metadata entries and 429 exported commands.

PR #1973, opened 2026-07-21
Test-MtLimitOnMicrosoftDomainUsage.ps1. Fix-licensing-and-mailbox-filter

### LINKED ISSUE #1840: 🪲 MT.1074 / Test-MtLimitOnMicrosoftDomainUsage: the no-Defender fallback always fails on the built-in Discovery Search Mailbox
### Pre-submission checklist

- [x] I can reproduce this issue using the latest released version of Maester (2.1.0).

### Describe the bug

`Test-MtLimitOnMicrosoftDomainUsage` (MT.1074) has two branches. With a Microsoft Defender XDR licence it queries outbound `EmailEvents` volume through Advanced Hunting, which is the right signal. Without Defender XDR it falls back to enumerating mailboxes and failing if any mailbox has a `.onmicrosoft.com` primary SMTP address.

The fallback itself is a sensible graceful degrade. The problem is it can never pass. Every Exchange Online tenant ships the built-in Discovery Search Mailbox (`DiscoverySearchMailbox{D919BA05-46A6-415f-80AD-7E09334BB852}`), its primary SMTP is permanently on `.onmicrosoft.com`, and Microsoft documents that it cannot be repurposed or converted. So a tenant that has done everything right, every user and shared mailbox on a custom domain, still fails MT.1074 on a system mailbox it is not allowed to change.

I did not believe this had gone unreported since the test merged in September, so I checked the existing issues (none match) and ran it live rather than assuming. It reproduces.

The match happens at line 69 of the function:

```powershell
$allMbx = Get-Mailbox | Where-Object { $_.PrimarySmtpAddress -like "*@*.onmicrosoft.com" }
```

`Get-Mailbox` returns the Discovery Search Mailbox by default, so it lands in `$allMbx`, the count is non-zero, and the test returns `$false`.

### Steps to reproduce

1. Use a tenant without a Defender XDR licence, so the function takes the `ExchangeOnline` fallback branch (`Get-MtLicenseInformation DefenderXDR` returns not-licensed).
2. Confirm every user and shared mailbox already uses a custom domain as its primary SMTP address, that is, the correct passing state.
3. `Connect-Maester`, then run the single test: `Invoke-Maester -Tag "MT.1074"`.
4. The test fails, listing exactly one mailbox: the Discovery Search Mailbox.

### Expected behavior

The test should pass when every mailbox an administrator can actually control uses a custom domain. The built-in Discovery Search Mailbox (`RecipientTypeDetails` of `DiscoveryMailbox`) should not count toward the failure, because it is a Microsoft system mailbox that cannot be moved off `.onmicrosoft.com`.

### Debug output

Reproduced live on a Business Basic tenant (no Defender XDR), delegated session, tenant domain redacted. The first block is the exact logic the test runs today, the second adds the proposed filter:

```text
=== CURRENT TEST LOGIC (line 69): Get-Mailbox | onmicrosoft primary ===
DisplayName              RecipientTypeDetails PrimarySmtpAddress
-----------              -------------------- ------------------
Discovery Search Mailbox DiscoveryMailbox     DiscoverySearchMailbox{D919BA05-46A6-415f-80AD-7E09334BB852}@<tenant>.onmicrosoft.com
MATCHES: 1   =>  MT.1074 would FAIL

=== WITH PROPOSED FILTER (-Filter "RecipientTypeDetails -ne 'DiscoveryMailbox'") ===
MATCHES: 0   =>  MT.1074 would PASS
```

### Module version

```text
Name    : Maester
Version : 2.1.0
```

### Environment data

```text
PSVersion   7.6.2
PSEdition   Core
OS          Microsoft Windows 10.0.26200
Platform    Win32NT
```

### Additional context

**Proposed fix.** One line, matching the exclusion filter Microsoft documents for this exact mailbox:

```powershell
$allMbx = Get-Mailbox -Filter "RecipientTypeDetails -ne 'DiscoveryMailbox'" | Where-Object { $_.PrimarySmtpAddress -like "*@*.onmicrosoft.com" }
```

The DefenderXDR branch is unaffected. This is only about the no-Defender fallback, which is otherwise a good graceful degrade. It just needs to skip the one mailbox no tenant can change.

**References.**

- Exchange team announcement that motivated the test: <https://techcommunity.microsoft.com/blog/exchange/limiting-onmicrosoft-domain-usage-for-sending-emails/4446167>
- Discovery Search Mailbox listed as a built-in system mailbox with that fixed GUID: <https://learn.microsoft.com/exchange/recipients-exchange-2013-help> (System mailboxes section)
- Discovery mailbox cannot be repurposed or converted, plus the documented `RecipientTypeDetails -eq/-ne 'DiscoveryMailbox'` filter syntax: <https://learn.microsoft.com/exchange/create-a-discovery-mailbox-exchange-2013-help>
- `Get-Mailbox` reference: <https://learn.microsoft.com/powershell/module/exchangepowershell/get-mailbox>
- Test introduced in PR #1113 (merged 2025-09-01).

JeanPhilippeGeorge:
## 📑 Test-MtLimitOnMicrosoftDomainUsage.ps1. Fix licensing and mailbox filter
+ Test for mdo (DefenderForOffice365P2) instead of DefenderXDR
+ In the ExchangeOnline part, filter out DiscoveryMailbox (as suggested by @Korthal-Maiyn) in issue #1840

Closes #1840

## ✅ Checks

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tenant capability detection for domain usage checks by using Microsoft 365 licensing details first, and only proceeding with Exchange Online checks after confirming connectivity.
  * Refined mailbox discovery to exclude discovery mailboxes and ensure primary addresses match the expected `*@*.onmicrosoft.com` pattern.
  * Strengthened Microsoft Graph user validation by only attempting lookups when a directory object ID is available, and reporting which matched mailboxes couldn’t be resolved in Graph.


merill:
Thank you for looking into this and fixing it @JeanPhilippeGeorge much appreciated. I made a small tweak to fix the CodeRabbit recommendation since the old code was only checking the first 1000 and would provide a Pass false positive.

merill (approved):
LGTM

JeanPhilippeGeorge:
@merill, nice catch. I didn't pay attention to the filtering on the client not on the server!

PR #1990, opened 2026-07-24
Suppress warnings for filtered NotRun tests

### LINKED ISSUE #1989: Suppress malformed-name warnings for filtered NotRun Pester tests
## Problem

A filtered Maester run emits thousands of misleading naming warnings for tests that were discovered by Pester but were never executed.

In [smoke run 30085246002](https://github.com/maester365/maester/actions/runs/30085246002), the workflow selected only `MT.1068`, while Pester discovered 12,988 tests. During conversion, `ConvertTo-MtMaesterResult` iterated over every discovered result and emitted 11,393 copies of the generic missing-colon warning in the Ubuntu job for filtered `NotRun` inner test names such as `Should pass all rules`.

## Root cause

`ConvertTo-MtMaesterResult` validates and warns about the `TestId: TestTitle` name format before handling `NotRun` results. Pester reports filtered tests as `NotRun`, including inner test names that intentionally do not carry a Maester test ID. The converter therefore warns for tests that were not selected or executed.

## Expected behavior

- Do not emit the missing-colon naming warning for filtered `NotRun` results.
- Continue converting and reporting `NotRun` results exactly as before.
- Continue emitting the warning when a malformed test actually executes (for example, a `Passed` or `Failed` result without `TestId: TestTitle`).
- Preserve existing IDs, titles, result details, counts, and report behavior.

## Acceptance criteria

- Regression coverage proves a malformed `NotRun` test produces no naming warning and remains present as a `NotRun` result.
- Regression coverage proves a malformed executed test still produces the naming warning.
- Focused `ConvertTo-MtMaesterResult` tests pass.

merill:
Closes #1989

## What changed

- Suppress the generic missing-colon naming warning only when Pester marks a discovered test as `NotRun`.
- Preserve conversion, IDs, titles, result details, counts, and report inclusion for those filtered tests.
- Add focused regression coverage proving filtered `NotRun` tests remain quiet while malformed executed tests still warn.

## Root cause

Pester discovers the full Maester suite before applying execution filters. `ConvertTo-MtMaesterResult` converted all discovered results and validated the `TestId: TestTitle` format before its `NotRun` handling. Inner test names such as `Should pass all rules` therefore produced one warning each even though they were filtered and never executed. In smoke run 30085246002 this generated 11,393 warnings while only `MT.1068` was selected.

## User impact

Filtered smoke and targeted runs no longer flood Actions logs with misleading naming warnings. Malformed tests that actually execute continue to surface the warning, so the validation signal is retained.

## Validation

- `Invoke-Pester -Path ./powershell/tests/functions/ConvertTo-MtMaesterResult.Tests.ps1 -Output Detailed` — 3 passed, 0 failed
- PowerShell parser checks for both changed files — passed
- `Invoke-ScriptAnalyzer` for both changed files — no findings
- `git diff --check` — passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Malformed test names no longer generate warnings when tests are filtered out and marked as **Not Run**.
  * Warnings continue to appear for executed tests with missing name delimiters.

* **Tests**
  * Added coverage to verify warning behavior and result handling for both skipped and executed tests.

### DIFF
--- a/powershell/internal/ConvertTo-MtMaesterResult.ps1
+++ b/powershell/internal/ConvertTo-MtMaesterResult.ps1
@@ -287,7 +287,7 @@
         if ($titleStart -gt 0) {
             $testId = $name.Substring(0, $titleStart).Trim()
             $testTitle = $name.Substring($titleStart + 1).Trim()
-        } else {
+        } elseif ($test.Result -ne 'NotRun') {
             Write-Warning "Test name does not contain a ':' character. Please use the format 'TestId: TestTitle' → $name"
         }
         $testResultDetail = $__MtSession.TestResultDetail[$test.ExpandedName]
--- a/powershell/tests/functions/ConvertTo-MtMaesterResult.Tests.ps1
+++ b/powershell/tests/functions/ConvertTo-MtMaesterResult.Tests.ps1
@@ -1,5 +1,50 @@
 BeforeAll {
     Import-Module "$PSScriptRoot/../../Maester.psd1" -Force
+
+    function Get-PesterResultsFixture {
+        param(
+            [Parameter(Mandatory = $true)]
+            [string] $TestName,
+
+            [Parameter(Mandatory = $true)]
+            [string] $Result
+        )
+
+        $duration = if ($Result -eq 'NotRun') { [TimeSpan]::Zero } else { [TimeSpan]::FromSeconds(1) }
+        $pesterTest = [PSCustomObject]@{
+            ExpandedName = $TestName
+            Result       = $Result
+            ErrorRecord  = @()
+            ScriptBlock  = [scriptblock]::Create('$true | Should -BeTrue')
+            Block        = [PSCustomObject]@{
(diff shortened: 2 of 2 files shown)


powershell/internal/ConvertTo-MtMaesterResult.ps1
```diff
@@ -287,7 +287,7 @@
         if ($titleStart -gt 0) {
             $testId = $name.Substring(0, $titleStart).Trim()
             $testTitle = $name.Substring($titleStart + 1).Trim()
-        } else {
+        } elseif ($test.Result -ne 'NotRun') {
```
SamErde, line 290: Just thinking "out loud" -- should this `if`/`else` block have a final `else` statement to cover edge cases as a best practice, or would that be unnecessary engineering?

merill: We are only skipping NotRun ones. Adding an else and not doing anthing there might be confusing.

PR #2001, opened 2026-07-25
fix(exchange): handle missing DKIM signing configurations

### LINKED ISSUE #1783: Test-MtCisDkim and Test-MtCisaDkim crash when accepted domain has no DkimSigningConfig
### Description

`Test-MtCisDkim` and `Test-MtCisaDkim` throw an unhandled error when an accepted domain
(typically `*.onmicrosoft.com`) has no corresponding `DkimSigningConfig` entry. The error
is caught by the function's `catch` block, which returns `$null` and reports the test as
`Ignored` with no useful diagnostic information.

### Steps to reproduce

1. Ensure your tenant has a `.onmicrosoft.com` accepted domain where `Get-DkimSigningConfig`
   does not return an entry (this is the default per
   [Microsoft documentation](https://learn.microsoft.com/defender-office-365/email-authentication-dkim-configure#use-exchange-online-powershell-to-configure-dkim-signing-of-outbound-messages))
2. Run `Test-MtCisDkim` or `Test-MtCisaDkim`

### Expected behavior

The test should either:
- Report a failure for the domain with a clear message ("No DkimSigningConfig found for
  {domain}"), or
- Skip the domain with a reason explaining that Microsoft auto-signs DKIM for
  `.onmicrosoft.com` domains even without an explicit config

### Actual behavior

```
Get-MailAuthenticationRecord: Cannot bind argument to parameter 'DkimDnsName' because it
is an empty string.
```

The `catch` block fires, `Add-MtTestResultDetail -SkippedBecause Error` is called, and the
function returns `$null`. The Pester test result is `Ignored` with no indication of the
underlying cause.

### Root cause

In `Test-MtCisDkim.ps1` (line 41) and `Test-MtCisaDkim.ps1` (line 40), when no
`DkimSigningConfig` matches the domain, `$dkimSigningConfig` is `$null`. The code then:

1. Accesses `$dkimSigningConfig.RotateOnDate` (null, date comparison falls through)
2. Sets `$Selector` to `$dkimSigningConfig.SelectorBeforeRotateOnDate` (null/empty)
3. For `.onmicrosoft.com` domains, evaluates `$dkimSigningConfig."$($selector)CNAME"` to
   empty string
4. Calls `Get-MailAuthenticationRecord -DkimDnsName ''` which throws

### Suggested fix

Add a null-check on `$dkimSigningConfig` at the start of the `foreach` loop body. If no
config exists, either fail the domain with a descriptive message or skip it with a reason.

### Environment

- Maester 2.1.0
- ExchangeOnlineManagement 3.9.2
- PowerShell 7.4
- Microsoft 365 Business Basic
- Tenant has two accepted domains: `contoso.com` (DKIM config exists) and
  `contoso.onmicrosoft.com` (no DKIM config, which is the Microsoft default)

### Workaround

Run `New-DkimSigningConfig -DomainName '<tenant>.onmicrosoft.com' -Enabled $true` to
explicitly create the DKIM signing config for the `.onmicrosoft.com` domain. This resolves
the null-reference because the config now exists in `Get-DkimSigningConfig` output.

amlhive-tech:
## Description

Closes #1783.

`Test-MtCisDkim` and `Test-MtCisaDkim` previously continued into selector and DNS processing when an accepted domain had no matching `DkimSigningConfig`. This caused an exception and left the Maester result without a useful domain-level outcome.

This update:

- skips domains where outbound sending is disabled;
- passes the tenant's `InitialDomain`, which Microsoft automatically DKIM-signs;
- fails active secondary `onmicrosoft.com`, hybrid, and custom domains when their DKIM configuration is missing;
- avoids selector rotation and DNS lookups when no signing configuration exists;
- adds focused regression coverage for every outcome in both DKIM checks; and
- adds Haris Habib (AMLHIVE) to the contributor registry.

## Impact

Tenants with an unlisted initial `onmicrosoft.com` domain no longer receive an ignored result caused by an exception. Other active domains are not incorrectly reported as DKIM compliant merely because their names end in `onmicrosoft.com`.

## Checks

- [x] Pester: 12 focused DKIM tests passed
- [x] PSScriptAnalyzer passed for the changed files
- [x] `git diff --check` passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved DKIM validation for accepted domains without matching signing configurations.
  * Reports parked domains as skipped, Microsoft-managed domains as passed, and other unmatched domains as failed with clear reasons.
  * Avoids unnecessary selector rotation and DNS checks when no signing configuration exists.
  * Provides consistent handling across CIS and CISA DKIM checks.

* **Tests**
  * Added coverage for DKIM outcomes across supported domain states.

* **Chores**
  * Added a new contributor to the project registry.

### DIFF
--- a/powershell/public/cis/Test-MtCisDkim.ps1
+++ b/powershell/public/cis/Test-MtCisDkim.ps1
@@ -1,4 +1,4 @@
-function Test-MtCisDkim {
+﻿function Test-MtCisDkim {
     <#
     .SYNOPSIS
     Checks state of DKIM for all EXO domains
@@ -45,8 +45,8 @@ function Test-MtCisDkim {
             if (-not $dkimSigningConfig) {
                 $dkimRecord = [PSCustomObject]@{
                     domain     = $domain.DomainName
-                    pass       = if ($domain.SendingFromDomainDisabled) { 'Skipped' } elseif ($domain.DomainName -like '*.onmicrosoft.com') { 'Passed' } else { 'Failed' }
-                    reason     = if ($domain.SendingFromDomainDisabled) { 'Parked domain' } elseif ($domain.DomainName -like '*.onmicrosoft.com') { 'Microsoft auto-signs DKIM for onmicrosoft.com domains' } else { 'No DkimSigningConfig found for domain' }
+                    pass       = if ($domain.SendingFromDomainDisabled) { 'Skipped' } elseif ($domain.InitialDomain -eq $true) { 'Passed' } else { 'Failed' }
+                    reason     = if ($domain.SendingFromDomainDisabled) { 'Parked domain' } elseif ($domain.InitialDomain -eq $true) { 'Microsoft auto-signs DKIM for the initial onmicrosoft.com domain' } else { 'No DkimSigningConfig found for domain' }
                     dkimRecord = $null
                 }
                 $dkimRecords += $dkimRecord
--- a/powershell/public/cis/Test-MtCisDkim.ps1
+++ b/powershell/public/cis/Test-MtCisDkim.ps1
@@ -1,4 +1,4 @@
-﻿function Test-MtCisDkim {
+function Test-MtCisDkim {
     <#
     .SYNOPSIS
     Checks state of DKIM for all EXO domains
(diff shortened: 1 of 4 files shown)


merill (changes requested):
Hi @amlhive-tech thanks for your PR!

Can I ask you split this into two PRs please? 

This will make it easier to track unrelated changes.

For the pester update we are currently pinned to 5.x and not Pester 6, so the rename of the tests as as will not be accurate. We might need to do this as a seperate PR that moves everything to Pester 6 in one go.

merill (approved):
LGTM tx for the fix.

PR #2048, opened 2026-07-28
fix: bound telemetry request timeout

### LINKED ISSUE #2047: Add a timeout to the Write-Telemetry HTTP request
## Problem

`Write-Telemetry` sends a non-critical PostHog event with `Invoke-RestMethod`, but the request does not set `-TimeoutSec`:

https://github.com/maester365/maester/blob/main/powershell/internal/Write-Telemetry.ps1

When `us.i.posthog.com` is blocked, silently dropped by a firewall/proxy, or otherwise unreachable, `Invoke-Maester` can pause until the operating system or HTTP stack's default timeout expires. Because telemetry runs synchronously before assessment execution, a best-effort telemetry call can materially delay or appear to hang an assessment.

## Reproduction

1. Block or blackhole outbound HTTPS traffic to `us.i.posthog.com` (a silent drop demonstrates the impact better than an immediate rejection).
2. Run `Invoke-Maester` without `-DisableTelemetry`.
3. Observe that execution pauses in `Write-Telemetry` before normal assessment work proceeds.

The exception is already caught, so the assessment eventually continues; the missing bound is the issue.

## Proposed change

Add an explicit, short timeout to the telemetry request, for example:

```powershell
Invoke-RestMethod ... -TimeoutSec 5 -ErrorAction Stop
```

Five seconds keeps telemetry best-effort and prevents an optional analytics endpoint from significantly delaying `Invoke-Maester`. A focused Pester test should verify that the bounded timeout is passed and that telemetry failures remain non-fatal.

SamErde:
## Summary

- bound the best-effort PostHog telemetry request to five seconds
- make the request error terminating so the existing catch reliably handles request failures
- add focused coverage for the timeout and non-fatal failure behavior

## Why

`Write-Telemetry` runs synchronously during `Invoke-Maester`. When the PostHog endpoint is blocked or unreachable, the request previously relied on the operating system or HTTP stack's default timeout and could materially delay an assessment. Telemetry is optional and should not hold up assessment execution.

Closes #2047.

## Validation

- PowerShell 7 / Pester 5.7.1: 2 passed, 0 failed
- Windows PowerShell / Pester 5.9.0: 2 passed, 0 failed
- `Build-MaesterModule.ps1`: passed
- `Test-MaesterModuleOutput.ps1`: passed
- `git diff --check`: passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Added a five-second timeout for telemetry requests.
  * Improved handling of telemetry network failures so they do not disrupt the application.

* **Tests**
  * Added automated coverage for request timeouts, error handling, and network failure scenarios.

### DIFF
--- a/powershell/internal/Write-Telemetry.ps1
+++ b/powershell/internal/Write-Telemetry.ps1
@@ -27,7 +27,7 @@
 
     # Send the POST request
     try {
-        Invoke-RestMethod -Uri $url -Method Post -ContentType "application/json" -Body $jsonBody | Out-Null
+        Invoke-RestMethod -Uri $url -Method Post -ContentType "application/json" -Body $jsonBody -TimeoutSec 5 -ErrorAction Stop | Out-Null
     }
     catch {
         Write-Verbose $_
--- a/powershell/tests/functions/Write-Telemetry.Tests.ps1
+++ b/powershell/tests/functions/Write-Telemetry.Tests.ps1
@@ -1,0 +1,27 @@
+Describe 'Write-Telemetry' {
+    BeforeAll {
+        . "$PSScriptRoot/../../internal/Write-Telemetry.ps1"
+    }
+
+    BeforeEach {
+        Mock Get-MgContext {
+            [PSCustomObject]@{ TenantId = '00000000-0000-0000-0000-000000000000' }
+        }
+    }
+
+    It 'uses a bounded timeout for the telemetry request' {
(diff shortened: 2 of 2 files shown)


merill (approved):
Thanks @SamErde this makes sense.

Long term I might look at using a proper OTel package for Maester.

Maybe https://opentelemetry.io/docs/zero-code/dotnet/getting-started/

This will allow users to configure where telemetry is sent to.

PR #2058, opened 2026-07-31
Fix BitLocker fixed drive encryption setting ID

### LINKED ISSUE #2015: 🪲 MT.1123: Fixed Encryption Type always shows 'Not configured' — settingDefinitionId typo (fde → fdv)
### Pre-submission checklist

- [x] I can reproduce this issue using the latest released version of Maester (2.1.0; also present on `main` as of 2026-07-27).

### Describe the bug

`Test-MtBitLockerFullDiskEncryption` (MT.1123) always reports **Fixed Encryption Type: Not configured** in its result table, even when the BitLocker policy explicitly enforces an encryption type for fixed data drives.

Cause: the child settingDefinitionId it matches against is misspelled at [`Test-MtBitLockerFullDiskEncryption.ps1` line 64 (tag 2.1.0)](https://github.com/maester365/maester/blob/2.1.0/powershell/public/maester/intune/Test-MtBitLockerFullDiskEncryption.ps1#L64):

```powershell
$fixedEncryptionTypeDropdownId = 'device_vendor_msft_bitlocker_fixeddrivesencryptiontype_fdeencryptiontypedropdown_name'
```

The real setting id uses **`fdv`** (Fixed Drive Volume — consistent with `fdvallowdra`, `fdvhiderecoverypage`, `fdvrecoverypasswordusagedropdown`, which this file spells correctly):

- `…_fixeddrivesencryptiontype_`**`fde`**`encryptiontypedropdown_name` → does **not** exist (`GET /beta/deviceManagement/configurationSettings/<id>` errors)
- `…_fixeddrivesencryptiontype_`**`fdv`**`encryptiontypedropdown_name` → exists ("Select the encryption type: (Device)")

So the `-eq` comparison in the child loop can never match, and `FixedEncryptionType` keeps its initialised `'Not configured'` for every tenant.

The pass/fail **verdict is unaffected** — `$hasFullEncryption` is set only in the OS-drive branch, which uses the correctly-spelled `osencryptiontypedropdown_name`. Only the evidence table is wrong.

### Steps to reproduce

1. Have an Endpoint Security > Disk Encryption BitLocker policy with **Enforce drive encryption type on fixed data drives** = Enabled and an encryption type selected (ours: Full encryption, i.e. `…fdvencryptiontypedropdown_name_1`).
2. Run `Test-MtBitLockerFullDiskEncryption` (or `Invoke-Maester` and view MT.1123's details).
3. The result table shows `Fixed Encryption Type: Not configured` regardless of the configured value.

Our observed output (the policy has enforced fixed-drive type = Full since a month before the run):

```
| Policy                   | Require Encryption | OS Encryption Type | Fixed Encryption Type | OS Cipher       |
| Enable BitLocker Profile | Enabled            | Used Space Only    | Not configured        | XTS-AES 256-bit |
```

### Expected behavior

`Fixed Encryption Type` should show the configured value (here **Full encryption**).

Suggested fix — one character class at line 64:

```powershell
$fixedEncryptionTypeDropdownId = 'device_vendor_msft_bitlocker_fixeddrivesencryptiontype_fdvencryptiontypedropdown_name'
```

### Debug output

Not captured — the defect is directly visible in the source vs the Graph settings catalog (links above).

### Module version

```
Maester 2.1.0 (pinned; installed from PSGallery in CI)
```

### Environment data

```
PSVersion  : 7.6.3 (Core)
Platform   : Unix — GitHub Actions ubuntu-latest (Ubuntu 24.04.4 LTS)
```

### Additional context

- Found while reconciling MT.1123's result table against the raw policy JSON from `GET /beta/deviceManagement/configurationPolicies/<id>/settings` — the table said "Not configured" while the policy carried `fdvencryptiontypedropdown_name_1`.
- Analysis was AI-assisted; every claim above was verified against the 2.1.0 tag source and live Graph metadata before filing.
- Happy to open a PR for the one-line fix if preferred.

massimomazzariol:
## Summary
- Correct the fixed drive encryption settingDefinitionId used by MT.1123
- Replace the invalid fde identifier segment with fdv

## Testing
- Parsed Test-MtBitLockerFullDiskEncryption.ps1 successfully with the PowerShell parser

Fixes #2015

### DIFF
--- a/powershell/public/maester/intune/Test-MtBitLockerFullDiskEncryption.ps1
+++ b/powershell/public/maester/intune/Test-MtBitLockerFullDiskEncryption.ps1
@@ -61,7 +61,7 @@
         $osEncryptionTypeId = 'device_vendor_msft_bitlocker_systemdrivesencryptiontype'
         $osEncryptionTypeDropdownId = 'device_vendor_msft_bitlocker_systemdrivesencryptiontype_osencryptiontypedropdown_name'
         $fixedEncryptionTypeId = 'device_vendor_msft_bitlocker_fixeddrivesencryptiontype'
-        $fixedEncryptionTypeDropdownId = 'device_vendor_msft_bitlocker_fixeddrivesencryptiontype_fdeencryptiontypedropdown_name'
+        $fixedEncryptionTypeDropdownId = 'device_vendor_msft_bitlocker_fixeddrivesencryptiontype_fdvencryptiontypedropdown_name'
         $requireEncryptionId = 'device_vendor_msft_bitlocker_requiredeviceencryption'
         $encryptionMethodId = 'device_vendor_msft_bitlocker_encryptionmethodbydrivetype'
         $osEncryptionMethodDropdownId = 'device_vendor_msft_bitlocker_encryptionmethodbydrivetype_encryptionmethodwithxtsosdropdown_name'


merill (approved):
LGTM 

Thanks for the fix @massimomazzariol

PR #2060, opened 2026-07-31
Fix Azure DevOps unavailable app skip status

### LINKED ISSUE #2051: 🪲 MT.1071 - Errors, but should skip if no Dev Ops App ID
### Pre-submission checklist

### Describe the bug

When I run Maester 2.2.11, and run as priviledged with an interactive login, for test MT.1071 I receive is, because Azure DevOps app is not available in this tenant. The issue is that its labeled as an "error" and not a "Skipped".

### Steps to reproduce

1. Connect-maester -service All -TenantId xxx -Privileged
2. invoke-maester -IncludeLongRunning -IncludePreview

### Expected behavior

That it should "Skipped" instead of "Error." 

It does show skipped in the output from the test, but in the report, it is categorized under "Error" instead of "Skipped."

### Debug output

```text

```

### Module version

```text
2.2.11
```

### Environment data

```text
PSVersion                      7.5.2
GitCommitId                    7.5.2
OS                             Darwin 25.5.0 Darwin Kernel Version 25.5.0: Mon…
Platform                       Unix
PSRemotingProtocolVersion      2.3
```

### Screenshots

Image

Image

### Additional context

_No response_

massimomazzariol:
## Summary
- Preserve the custom skipped status when the Azure DevOps service principal is unavailable
- Limit Graph request error handling so Pester skip control flow is not converted into an error
- Add regression coverage for skip propagation

## Testing
- Reproduced the skip conversion behavior with Pester 5.7.1
- Verified the regression test failed before the fix and passed afterward
- Passed all 4 Test-MtCaAzureDevOps tests with Pester 5.7.1
- Parsed both modified PowerShell files successfully
- Ran git diff --check successfully

Fixes #2051

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Azure DevOps service-principal lookup error handling.
  * Missing service principals are now correctly reported as skipped before policy evaluation.
  * Custom skip results now remain skips instead of being incorrectly recorded as errors.

### DIFF
--- a/powershell/public/maester/entra/Test-MtCaAzureDevOps.ps1
+++ b/powershell/public/maester/entra/Test-MtCaAzureDevOps.ps1
@@ -33,10 +33,17 @@
 
     try {
         $azureDevOpsServicePrincipal = Invoke-MtGraphRequest -RelativeUri 'servicePrincipals' -ApiVersion v1.0 -Filter "appId eq '$azureDevOpsAppId'" -Select id
-        if (-not $azureDevOpsServicePrincipal) {
-            Add-MtTestResultDetail -SkippedBecause Custom -SkippedCustomReason "Azure DevOps app (App ID: $azureDevOpsAppId) is not available in this tenant."
-            return $null
-        }
+    } catch {
+        Add-MtTestResultDetail -SkippedBecause Error -SkippedError $_
+        return $null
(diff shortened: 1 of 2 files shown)


merill (approved):
LGTM

PR #2143, opened 2026-08-22
(MT.1178) - ASR rules in setting catalog policies were not detected

royklo:
## Problem

`Test-MtIntuneASRRules` filters policies on `templateReference/templateFamily eq 'endpointSecurityAttackSurfaceReduction'`, which only matches ASR profiles created under **Endpoint security > Attack surface reduction**.

ASR rules created via **Devices > Configuration > Settings catalog** land on a policy with `templateFamily: "none"` but use the *identical* setting definition IDs — the definitions are published with `visibility: "settingsCatalog,template"`, so it's one shared definition surfaced in both places.

Result: a tenant with all three Standard Protection baseline rules on **Block** via the Settings catalog reports *"No Endpoint Security Attack Surface Reduction policies found"* and MT.1178 fails. Verified against such a tenant — all 17 configured rules were invisible to the current query.

Related: **Device Control** and **Exploit Protection** profiles also carry the `endpointSecurityAttackSurfaceReduction` family, so they inflated the policy count and produced empty `0 rules` tables.

## Fix

- Query Windows configuration policies (`$filter=platforms has 'windows10'`) and identify ASR policies by the settings they contain, not by how they were authored.
- Skip policies with no ASR rules — this drops Device Control and Exploit Protection automatically.
- Move the "nothing found" branch downstream so it means *no ASR rules anywhere* rather than *no ES-family policies*.
- Report now shows whether each policy came from **Endpoint Security** or the **Settings catalog**.

CleanShot 2026-08-22 at 20 30 41

## Testing

New `powershell/tests/functions/Test-MtIntuneASRRules.Tests.ps1`, 5/5 passing: settings-catalog detection (the regression), Device Control exclusion, baseline rule not in Block/Audit → `$false`, no ASR rules → `$false`, unlicensed → skipped. PSScriptAnalyzer clean.

## Docs

Companion `.md` updated to note both authoring surfaces count. Generated pages (`MT.1178.md`, the command `.mdx`) left untouched — bots refresh those on push to `main`.

## Known gap

ADMX-backed `ExploitGuard_ASR_Rules` stores rules as a GUID plus numeric state (`0`/`1`/`2`/`6`) instead of a `_block`/`_audit` suffix, so those still parse as *Not configured*. Left out rather than shipping untested parsing — happy to follow up.

## Summary by CodeRabbit

* **New Features**
  * ASR rule checks now include Intune Endpoint security profiles and Settings catalog policies.
  * Results identify each policy’s source and exclude policies without ASR rules.

* **Bug Fixes**
  * Standard Protection rules must be set to Block or Audit; Warn no longer passes.
  * Improved reporting when no ASR configuration exists and updated compliance counts.

* **Documentation**
  * Clarified supported policy types, rule modes, and evaluation requirements.

### DIFF
--- a/powershell/public/maester/intune/Test-MtIntuneASRRules.md
+++ b/powershell/public/maester/intune/Test-MtIntuneASRRules.md
@@ -27,6 +27,13 @@ See the [Microsoft Defender ASR rules deployment guide](https://learn.microsoft.
 
 Additional ASR rules detected in tenant policies are reported for visibility but do not affect the pass/fail result. **Warn** is a supported ASR rule state but does not satisfy the baseline. Baseline rules in **Audit** mode will trigger an informational note recommending a transition to **Block** mode.
 
+ASR rules can be configured from two places in Intune, and both count towards this check:
+
+- **Endpoint security** > **Attack surface reduction** (the Attack Surface Reduction Rules profile)
+- **Devices** > **Configuration** > **Settings catalog** (the **Defender** > **Attack Surface Reduction Rules** category)
+
+Both surfaces write the same underlying Defender settings, so the check evaluates every Windows configuration policy and reports which surface each policy was authored from.
+
 #### Remediation action
 
 1. Navigate to [Microsoft Intune admin center](https://intune.microsoft.com).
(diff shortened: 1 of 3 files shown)


merill (approved):
Cheers. Thanks a lot for the fixes @royklo

PR #2144, opened 2026-08-22
MT.1123: detect BitLocker configured via Settings catalog

royklo:
Same root cause as #2143, different test.

## Problem

`Test-MtBitLockerFullDiskEncryption` filters policies on `templateReference/templateFamily eq 'endpointSecurityDiskEncryption'`, which only matches profiles created under **Endpoint security > Disk encryption**.

Every `device_vendor_msft_bitlocker_*` setting definition is published with `visibility: "settingsCatalog,template"` — including the four this test reads (`systemdrivesencryptiontype`, `fixeddrivesencryptiontype`, `requiredeviceencryption`, `encryptionmethodbydrivetype`). BitLocker configured through **Devices > Configuration > Settings catalog** therefore lands on `templateFamily: "none"` and is invisible to the test.

Verified against a tenant holding a `CIS (BL) BitLocker - Windows 11 Intune 5.0.0` settings-catalog policy with 7 BitLocker settings configured. MT.1123 reported:

> No Endpoint Security Disk Encryption policies found in Intune. Create a Disk Encryption policy under **Endpoint Security > Disk Encryption**…

That policy exists. It just has **Used Space Only** on OS drives — so the check still fails, but for the wrong reason and with remediation advice that sends the admin to create a duplicate policy instead of changing one dropdown in the one they have. In any tenant where a settings-catalog BitLocker policy *does* set Full encryption, it's a straight false negative.

Related: **Personal Data Encryption** profiles also carry the `endpointSecurityDiskEncryption` family, so they counted towards the Disk Encryption policy total despite configuring no BitLocker settings.

## Fix

- Query Windows configuration policies (`$filter=platforms has 'windows10'`) and identify BitLocker policies by the settings they contain, not by template family.
- Skip policies with no BitLocker settings — drops Personal Data Encryption automatically.
- Add the missing `Test-MtConnection Graph` check, which this helper lacked.
- Report shows whether each policy was configured via **Endpoint Security** or the **Settings catalog**.

Trade-off: one `/settings` call per Windows policy. Graph has no server-side filter on setting definition IDs, so this is unavoidable for correctness; `platforms has 'windows10'` narrows the set and the `Invoke-MtGraphRequest` session cache absorbs repeats.

## Testing

New `powershell/tests/functions/Test-MtBitLockerFullDiskEncryption.Tests.ps1`, 6/6 passing:

- BitLocker full encryption in a Settings catalog policy → `$true` (the regression)
- Personal Data Encryption profile excluded from the report
- Used Space Only → `$false`, with the policy still named in the output rather than claiming none exists
- No BitLocker settings anywhere → `$false`
- Graph not connected → skipped
- Intune not licensed → skipped

PSScriptAnalyzer clean.

## Docs

Companion `.md` updated: notes that both authoring surfaces count, and the rationale section was tightened. Generated pages (`MT.1123.md`, the command `.mdx`) left untouched — `update-test-docs.yaml` and `update-module-docs.yaml` refresh those on push to `main`.

One thing worth deciding before you open it: the test still gates only on the OS drive. That CIS policy has FixedEncryptType: Not configured, which stays unflagged even now that the policy is visible. That's existing design, not a regression, so I left it out of scope — but it may be worth raising in the PR thread or as a separate issue.

## Summary by CodeRabbit

* **New Features**
  * BitLocker full-disk encryption checks now cover configurations from both Endpoint security and Settings catalog policies.
  * Reports identify the configuration source and flag operating system drives configured for Used Space Only encryption.

* **Bug Fixes**
  * Improved handling when no applicable BitLocker settings are found.
  * Added clearer outcomes for unavailable Graph connectivity or Intune licensing.

* **Documentation**
  * Clarified full encryption requirements, risks of Used Space Only encryption, and supported configuration locations.

### DIFF
--- a/powershell/public/maester/intune/Test-MtBitLockerFullDiskEncryption.md
+++ b/powershell/public/maester/intune/Test-MtBitLockerFullDiskEncryption.md
@@ -1,14 +1,13 @@
-Ensure at least one Intune Disk Encryption policy enforces BitLocker with **full disk encryption type**.
+Ensure at least one Intune policy enforces BitLocker **full encryption** on operating system drives.
 
-BitLocker Drive Encryption protects data on Windows devices by encrypting the disk.
-However, BitLocker supports two encryption types that have very different security implications:
+BitLocker offers two encryption types, and the difference matters:
 
-- **Full disk encryption** — encrypts the entire drive including free space. This is the recommended and secure option.
-- **Used space only encryption** — only encrypts sectors currently holding data. **This is dangerous on drives that previously contained unencrypted data**, because previously deleted files remain as raw data in unencrypted free space. This data can be recovered using commonly available data recovery software (e.g., Recuva, PhotoRec, or forensic imaging tools). NTFS marks deleted file sectors as "free" but does not zero them out — the original bytes stay on disk until overwritten by new data.
+- **Full encryption** encrypts the entire drive, including free space.
+- **Used Space Only encryption** encrypts only the sectors that currently hold data.
 
-**Bottom line:** If BitLocker is enabled with "Used space only" on a drive that already had data on it before encryption was turned on, that pre-existing deleted data is fully recoverable. Only "Full disk encryption" guarantees that the entire drive surface is protected.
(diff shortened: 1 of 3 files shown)


merill (approved):
Love it. This is a good fix to include. Tx @royklo

PR #2147, opened 2026-08-24
feat(intune): add macOS compliance and enrollment checks (MT.1214-MT.1217)

DanielLystad:
## 📑 Description

Adds four macOS checks from the Intune test spec list:

| ID | Check | Severity |
| --- | --- | --- |
| MT.1214 | macOS compliance policy requires System Integrity Protection | Medium |
| MT.1215 | Gatekeeper restricts app download locations | Medium |
| MT.1216 | macOS compliance policy requires a Defender machine risk score level | Medium |
| MT.1217 | macOS LAPS is configured on Automated Device Enrollment profiles | High |

Apple coverage today is the four token-expiry checks (MT.1092-MT.1095). Nothing reads `deviceCompliancePolicies` outside an internal Defender helper, and nothing reads `depOnboardingSettings` at all, so there is no overlap with the Zero Trust assessment port.

Note that SIP cannot be enforced by MDM — it is toggled from Recovery on the device — so MT.1214 checks that it is *required* as a compliance condition, which is what makes a Mac with SIP disabled non-compliant and blockable by Conditional Access. macOS LAPS is likewise not Windows LAPS: it lives on the ADE enrollment profile rather than in a settings catalog policy, which is why MT.1217 needs a different endpoint.

- MT.1214-MT.1216 share `Get-MtMacOSCompliancePolicy`; MT.1217 uses `Get-MtMacOSEnrollmentProfile` for the two-hop `depOnboardingSettings` → `enrollmentProfiles` query
- All four require the policy or profile to be **assigned** — an unassigned compliance policy is never evaluated against a device
- `Get-MtMacOSEnrollmentProfile` never retains `adminAccountPassword`, which `depMacOSEnrollmentProfile` exposes as a string; there is a unit test asserting it cannot reach a result
- MT.1217 returns **Investigate** rather than a pass when a compliant profile coexists with one that provisions an admin account and no rotation, since that second profile hands every device a static local admin password. Follows the pattern in MT.1196
- A failed Graph call arrives as a non-terminating error with a null result, which collects as one element with no properties. Both helpers guard on `id` and return `$null` so the caller reports `NotAuthorized` rather than "no policies found" — `depOnboardingSettings` is gated by Intune RBAC on top of the Graph scope and 403s for accounts without enrollment programs access
- Unrecognised enum values fall through to their raw form rather than being labelled "Not configured"
- `MacOS` casing follows the Graph resources being queried; happy to change to `MacOs` if you prefer
- **No new Graph scopes** — `DeviceManagementConfiguration.Read.All` and `DeviceManagementServiceConfig.Read.All` are already in `Get-MtGraphScope`
- Test numbers reserved in #697

## ✅ Checks

- [x] My code requires changes to the documentation.
- [x] I have updated the documentation as required.

## ℹ️ Additional Information

Verified on macOS with PowerShell 7.6.1:

- 41 unit tests in `powershell/tests/functions/Test-MtMacOSSecurity.Tests.ps1`, covering the pass, fail, unassigned-policy, empty-tenant, unrecognised-enum and not-authorized branches for all four checks and both helpers
- `./powershell/tests/pester.ps1` — 10491 tests, no new failures against the `main` baseline
- PSScriptAnalyzer clean at Error, Warning and Information level

Verified against a live tenant, both outcomes rather than only the passing path: MT.1214 and MT.1215 passed, MT.1216 failed on a risk score of `unavailable`, and MT.1217 returned Investigate across three ADE profiles — one with a rotated admin account, one with an admin account and no rotation, one with neither.

Reading `depOnboardingSettings` needed more than Intune Administrator in that tenant: the Graph scope was present and the call still returned 403 until a higher role was active. That is why MT.1217 maps 403 to `NotAuthorized` instead of failing.

## Summary by CodeRabbit

* **New Features**
  * Added macOS Intune checks for System Integrity Protection, Gatekeeper, Defender risk scores, and LAPS configuration.
  * Added support for evaluating compliance and configuration policies, enrollment profiles, assignments, and security settings.
  * Added four macOS security checks to the default test configuration.

* **Documentation**
  * Added guidance on prerequisites, security impact, remediation, configuration, and Microsoft references.

* **Tests**
  * Added automated coverage for policy evaluation, enrollment profiles, authorization handling, edge cases, and security findings.


DanielLystad:
Thanks — both addressed in adbeb9f.

**Enrollment profile request (major):** confirmed, and a real gap. The `depOnboardingSettings` request was guarded but the per-token `enrollmentProfiles` request was not, so a failure there was filtered out by the id check and left a partial or empty list. `Test-MtMacOSLAPSConfiguration` would then report a failed control rather than a skip. Applied the same guard.

Two regression tests cover it: a valid token whose profile request fails must return `$null`, and a token holding only iOS profiles must still return an empty list so a genuine no-macOS-profiles tenant is not misread as a failure.

Checking that also turned up a weakness in my own assertions: `Should -BeNullOrEmpty` is satisfied by an empty collection as well as by `$null`, so it could not distinguish the two states those tests exist to separate. Switched them to `$null -eq …`. Reverting the fix now produces exactly one failure, on the intended test.

**Apple support URLs (nitpick):** applied. Both forms resolve, but the numeric IDs are the current canonical ones.

43 unit tests passing, PSScriptAnalyzer clean, no new failures in `./powershell/tests/pester.ps1` against the `main` baseline.

DanielLystad:
Rebased on `main` and extended MT.1215 in e42f2f1, prompted by #2143 and #2144.

Those two broadened MT.1178 and MT.1123 to detect policies authored in the settings catalog. MT.1215 had the same gap in the other direction: it only read the macOS compliance policy, so a tenant enforcing the `com.apple.systempolicy.control` payload without a Gatekeeper compliance rule failed the check despite genuinely restricting app sources. Since the macOS endpoint protection template is deprecated, settings catalog is where new Gatekeeper configuration gets authored, so that will be the common case over time.

The check now passes if an assigned policy of either kind restricts app sources, and the result table carries a `Source` column distinguishing them, matching how #2144 presents it. A policy that disables Gatekeeper assessment does not count. Renamed to "Ensure Gatekeeper restricts macOS app download locations" since it is no longer compliance-only.

Worth noting the two mechanisms are not equivalent — a compliance policy observes and feeds Conditional Access, a configuration policy enforces — but both restrict app sources, which is what the check is named for. Happy to split them into separate checks instead if you would rather keep detection and enforcement distinct.

Verified against a tenant holding both:

| Policy | Source | Allowed app source | Assignments |
| --- | --- | --- | --- |
| Mac Compliance | Compliance policy | App Store and identified developers | 1 |
| Gatekeeper | Settings catalog | Enforced: App Store and identified developers | 3 |

49 unit tests, PSScriptAnalyzer clean, no new failures in `./powershell/tests/pester.ps1` against the `main` baseline.

DanielLystad:
Both addressed in 42467ff.

**Per-policy request guards:** correct, and the same gap I had already fixed on the enrollment helper. The policy list request was guarded but the per-policy `settings` and `assignments` requests were not. A failed settings read left no setting instances, so the policy was skipped silently and an enforced tenant looked unconfigured. Both now apply the guard.

**Absent `allowidentifieddevelopers` key:** also correct, and the more interesting of the two. Coercing it to `$false` reported "Enforced: Mac App Store only", which claims a *stricter* posture than the policy actually enforces — an MDM payload only applies the keys it contains, so an omitted key leaves the device's existing setting untouched. It is now `$null` and renders as "assessment on, identified developers not specified".

Reverting either fix produces exactly two failures on the intended tests.

Checking those also turned up two problems in the new tests themselves, worth noting since both would have quietly weakened coverage:

- The `Describe`-level default mock for `Get-MtMacOSGatekeeperEnforcement` shadowed the real function in that helper's own tests, so the helper tests were exercising the mock. Scoped it to the contexts that exercise MT.1215.
- A mock filter used `'*configurationPolicies?*'`, and `?` is a single-character wildcard in `-like`, so it also matched the per-policy `settings` and `assignments` URIs.

52 unit tests, PSScriptAnalyzer clean, no new failures against the `main` baseline.

merill (approved):
LGTM

Issue #2105, opened 2026-08-06
🪲 MT.1011 - false positive result, failed although CA exists


blindzero:
### Pre-submission checklist

- [x] I can reproduce this issue using the latest released version of Maester.
- [x] I have searched the existing issues and did not find a similar report.

### Describe the bug

Test **MT.1011: At least one Conditional Access policy is configured to secure security info registration only from a trusted location.** is failing with result: `No conditional access policy securing security info registration.`

This is a false positive for our environment, see screenshot below.

### Steps to reproduce

1. Run MT.1011

### Expected behavior

Test Success

### Module version

```text
ModuleType Version    PreRelease Name                                PSEdition ExportedCommands
---------- -------    ---------- ----                                --------- ----------------
Script     2.2.0                 Maester                             Core,Desk {Add-MtMaesterAppFederatedCredential, A…
```

### Environment data

```text
Name                           Value
----                           -----
PSVersion                      7.6.4
PSEdition                      Core
GitCommitId                    7.6.4
OS                             Microsoft Windows 10.0.26200
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.4
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

### Screenshots

<img width="473" height="652" alt="Image" src="https://github.com/user-attachments/assets/fb0e6f0a-dee8-47a2-adb3-0ebda257c716" />

### Additional context

MT.1011 is only considering policies if

* **All** Users included _and_
* **All** Client App Types included _and_
* **All** locations are included _and_
* **No** locations are excluded

Although I get the idea of that `registersecurityinfo` the error output is misleading as it says: 

> "No Conditional Access policy securing security info registration."

Additionally, synopsis says

> Checks if the tenant has at least one Conditional Access policy securing security info registration.

Minimum: making clear All apps and locations must be included with no locations excluded.
If I assume that Synopsis and error output, it should be sufficient to have _any_ CA on `urn:user:registersecurityinfo`.

PR #2165, opened 2026-09-01
fix(MT.1191): skip when break-glass exclusion cannot be verified

### LINKED ISSUE #2114: MT.1191 reports Passed when the break-glass exclusion cannot be verified
### Pre-submission checklist

### Describe the bug

MT.1191 is reported as Passed when an enabled Compliant Network enforcement policy exists but no emergency access accounts are configured in `maester-config.json`.
In this situation, `Test-MtGsaCompliantNetworkBreakGlassExcluded` records the following result:
> No emergency access accounts are configured in maester-config.json (EmergencyAccessAccounts), so break-glass exclusion cannot be verified.
The function then returns `$null`, without setting `-SkippedBecause` or `-Investigate`.
The bundled Pester wrapper only performs an assertion when the result is non-null:
```powershell
$result = Test-MtGsaCompliantNetworkBreakGlassExcluded

if ($null -ne $result) {
    $result | Should -Be $true -Because "a Compliant Network block policy must never lock out emergency access accounts."
}
```
No assertion is executed, so the control is reported as Passed.
A control that explicitly states the required exclusion could not be verified is therefore included in the Passed count.

### Steps to reproduce

This can be reproduced locally without a Global Secure Access license, by mocking an applicable policy and an empty emergency access account configuration.
1. The function returns null while reporting that the exclusion cannot be verified
```powershell
Describe "MT.1191 function repro" {
    It "returns null while reporting that the exclusion cannot be verified" {
        InModuleScope Maester {
            Mock Test-MtConnection { $true }
            Mock Get-MtLicenseInformation { 'P1' }
            Mock Get-MtCompliantNetworkPolicy {
                @([pscustomobject]@{ displayName = 'LAB Compliant Network'; id = 'lab-policy-01' })
            }
            Mock Get-MtEmergencyAccessAccount { @() }
            Mock Add-MtTestResultDetail { }

            $result = Test-MtGsaCompliantNetworkBreakGlassExcluded

            $result | Should -BeNullOrEmpty
            Should -Invoke Add-MtTestResultDetail -Times 1 -ParameterFilter {
                $Result -like '*cannot be verified*'
            }
        }
    }
}
```
Reproduction test result: `Tests Passed: 1, Failed: 0`
2. The bundled wrapper logic reports Passed
The dependencies are mocked only to reach the relevant branch.
The assertion block below is copied unchanged from the bundled wrapper:
```powershell
Describe "MT.1191 wrapper repro" {
    It "MT.1191: Break-glass accounts should be excluded from the Compliant Network Conditional Access policy." {
        InModuleScope Maester {
            Mock Test-MtConnection { $true }
            Mock Get-MtLicenseInformation { 'P1' }
            Mock Get-MtCompliantNetworkPolicy {
                @([pscustomobject]@{ displayName = 'LAB Compliant Network'; id = 'lab-policy-01' })
            }
            Mock Get-MtEmergencyAccessAccount { @() }

            $result = Test-MtGsaCompliantNetworkBreakGlassExcluded

            if ($null -ne $result) {
                $result | Should -Be $true -Because "a Compliant Network block policy must never lock out emergency access accounts."
            }
        }
    }
}
```
Wrapper reproduction result: `Tests Passed: 1, Failed: 0`
The published wrapper logic completes as Passed even though the function returned `$null` after recording that the exclusion could not be verified.

### Expected behavior

A control whose required reference configuration is absent, and whose own result states that the exclusion cannot be verified, should not be reported as Passed.
The appropriate disposition is a design decision for the maintainers. In the current result model, `Investigate` or an explicit skipped reason may be options. A future result model could represent this as an indeterminate outcome.
To be clear about the scope of this report: I am not suggesting that an empty `EmergencyAccessAccounts` configuration should in itself be a security failure. The issue is that the exclusion was not evaluated, while the control is counted as successfully evaluated.

### Debug output

```text

```

### Module version

```text
Maester 2.2.0
```

### Environment data

```text
PowerShell 7.6.4
Pester 5.9.0
Windows 11 25H2
```

### Screenshots

Image

### Additional context

A second branch of the same function follows the same null-return pattern, when no enabled Compliant Network enforcement policy exists:
> No enabled Compliant Network enforcement policy was found, so there is nothing to evaluate.
I observed this branch in an actual run against a tenant with no Compliant Network policy. The HTML report displayed the control with a green Passed badge directly above that sentence, and the run summary reported:
```
Passed: 1, Failed: 0, Investigate: 0, Skipped: 0
```
Screenshot attached.
That branch is arguably non-applicability rather than an unverifiable exclusion, so it may warrant a different disposition. Both branches currently collapse into Passed because they return `$null` without declaring another test status.

earbona23:
Fixes #2114.

## The problem

MT.1191 is counted as **Passed** in exactly the situation where it has just reported that it could not check anything.

When no emergency access accounts are configured in `maester-config.json`, `Test-MtGsaCompliantNetworkBreakGlassExcluded` records *"break-glass exclusion cannot be verified"* and returns `$null`. The bundled wrapper only asserts when the result is non-null:

```powershell
$result = Test-MtGsaCompliantNetworkBreakGlassExcluded

if ($null -ne $result) {
    $result | Should -Be $true -Because "..."
}
```

So the `It` block completes without executing a single assertion, and Pester reports Passed. A control that states its own requirement is unverifiable lands in the passed count — which is the reading most likely to stop someone from looking.

## The fix

`Add-MtTestResultDetail` calls `Set-ItResult -Skipped` only when `-SkippedBecause` is supplied. Passing `-SkippedBecause Custom` with the same message keeps the wording identical and moves the result from Passed to Skipped:

```diff
-Add-MtTestResultDetail -Result No emergency access accounts are configured...
+Add-MtTestResultDetail -SkippedBecause Custom -SkippedCustomReason No emergency access accounts are configured...
```

That pattern is already used in 153 places in this repository, so this introduces nothing new.

## Tests

Adds `powershell/tests/functions/Test-MtGsaCompliantNetworkBreakGlassExcluded.Tests.ps1` covering three branches: the unverifiable case now skips, an unexcluded break-glass account still returns `$false`, and an excluded one still returns `$true`.

The first test was checked by **reverting the one-line fix and confirming it fails** (`Expected Add-MtTestResultDetail to be called 1 time exactly, but was called 0 times`). Without that check it would pass with or without the bug and prove nothing.

Full unit suite after the change: **4103 passed, 0 failed** (Pester 6.1.0, pwsh 7.6.2, Linux).

## Deliberately not changed

The sibling `"No enabled Compliant Network enforcement policy was found"` branch, and the equivalent branches in the other GSA checks, are left alone. Those are genuinely **not applicable** rather than **unverifiable** — with no such policy there is no lock-out risk — and they follow the convention established across the GSA family. Happy to revisit separately if you see it differently.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated compliance results to clearly show when verification is skipped because no emergency access accounts are configured.
  * Prevented unverifiable controls from being incorrectly counted as passed.
  * Preserved accurate pass/fail results for network policies that do or do not exclude break-glass accounts.

* **Tests**
  * Added coverage for skipped, compliant, and non-compliant scenarios.

### DIFF
--- a/powershell/public/maester/globalsecureaccess/Test-MtGsaCompliantNetworkBreakGlassExcluded.ps1
+++ b/powershell/public/maester/globalsecureaccess/Test-MtGsaCompliantNetworkBreakGlassExcluded.ps1
@@ -45,7 +45,7 @@
 
         $emergencyAccounts = Get-MtEmergencyAccessAccount
         if (-not $emergencyAccounts) {
-            Add-MtTestResultDetail -Result 'No emergency access accounts are configured in maester-config.json (EmergencyAccessAccounts), so break-glass exclusion cannot be verified.'
+            Add-MtTestResultDetail -SkippedBecause Custom -SkippedCustomReason 'No emergency access accounts are configured in maester-config.json (EmergencyAccessAccounts), so break-glass exclusion cannot be verified.'
             return $null
         }
 
--- a/powershell/tests/functions/Test-MtGsaCompliantNetworkBreakGlassExcluded.Tests.ps1
+++ b/powershell/tests/functions/Test-MtGsaCompliantNetworkBreakGlassExcluded.Tests.ps1
@@ -1,0 +1,76 @@
+﻿BeforeAll {
+    Import-Module "$PSScriptRoot/../../Maester.psd1" -Force
+}
+
+Describe 'Test-MtGsaCompliantNetworkBreakGlassExcluded' {
+    Context 'When no emergency access accounts are configured' {
+        It 'Skips the test instead of letting it be counted as passed' {
+            # The check states that break-glass exclusion cannot be verified and returns
+            # $null. The bundled MT.1191 wrapper only asserts when the result is non-null,
+            # so without a skip reason the It block completes with no assertion at all and
+            # Pester reports Passed. Add-MtTestResultDetail calls Set-ItResult -Skipped only
+            # when -SkippedBecause is supplied, which is what keeps an unverifiable control
+            # out of the passed count.
+            InModuleScope Maester {
+                Mock Test-MtConnection { $true }
+                Mock Get-MtLicenseInformation { 'P1' }
+                Mock Get-MtCompliantNetworkPolicy {
+                    @([pscustomobject]@{ displayName = 'Compliant Network'; id = 'policy-01' })
+                }
+                Mock Get-MtEmergencyAccessAccount { @() }
+                Mock Add-MtTestResultDetail { }
+
+                $result = Test-MtGsaCompliantNetworkBreakGlassExcluded
+
+                $result | Should -BeNullOrEmpty
+                Should -Invoke Add-MtTestResultDetail -Times 1 -Exactly -ParameterFilter {
+                    $SkippedBecause -eq 'Custom' -and $SkippedCustomReason -like '*cannot be verified*'
+                }
+            }
+        }
+    }
+
+    Context 'When emergency access accounts are configured' {
+        It 'Returns false when a policy does not exclude a break-glass account' {
+            InModuleScope Maester {
+                Mock Test-MtConnection { $true }
+                Mock Get-MtLicenseInformation { 'P1' }
+                Mock Get-MtCompliantNetworkPolicy {
+                    @([pscustomobject]@{
+                            displayName = 'Compliant Network'
+                            id          = 'policy-01'
+                            conditions  = @{ users = @{ excludeUsers = @(); excludeGroups = @() } }
+                        })
+                }
+                Mock Get-MtEmergencyAccessAccount {
+                    @([pscustomobject]@{ ObjectId = 'bg-user-01'; DisplayName = 'Break Glass 01'; Type = 'user' })
+                }
+                Mock Invoke-MtGraphRequest { @() }
+                Mock Add-MtTestResultDetail { }
+
+                Test-MtGsaCompliantNetworkBreakGlassExcluded | Should -BeFalse
+            }
+        }
+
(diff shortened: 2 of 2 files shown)


earbona23:
Thanks — automated checks are green (CodeRabbit + Codacy). Happy to rebase onto the latest `main` whenever it helps the review; it still merges cleanly today. Just flagging I'm around to adjust anything.

For context: the change is a one-liner that follows the `-SkippedBecause Custom` pattern already used in 153 places here, and the new test was verified by reverting the fix and confirming it fails (so it can't pass without the fix). Full unit suite after the change: 4103 passed, 0 failed. If a maintainer sees the "not applicable vs unverifiable" split on the sibling GSA branches differently, glad to revisit.

earbona23:
Bumping this one because it's the oldest of a small batch, and they're easier to triage together than one at a time.

Four MT.* correctness fixes are open, all green on CodeRabbit and Codacy, all with tests verified by reverting the fix and confirming the new test fails:

- #2165 MT.1191 — reports PASS when the break-glass exclusion cannot be verified; now skips with `-SkippedBecause Custom`, the pattern already used in 153 places here.
- #2173 MT.1024 — configured break-glass accounts are not excluded from the sign-in/user risk recommendation.
- #2174 MT.1107-1109 — Graph errors are indistinguishable from an empty result; now surfaced as Skipped instead of a silent pass.
- #2175 MT.1011 — a browser-scoped security info registration policy is a false negative (#2105).

They share a theme: a check that cannot see the data should not report that the data is fine. Each is independent and rebases cleanly; happy to split, squash or reorder them however suits your review flow.

merill (approved):
LGTM

PR #2166, opened 2026-09-02
feat(website): add pronunciation audio to homepage hero

BakkerJan:
## 📑 Description

"Maester" gets pronounced three or four different ways in meetings, on podcasts and at conferences, and people who are unsure of a name tend to avoid saying it at all. This adds a small speaker button on the corner of the Maester logo in the homepage hero — click it and you hear how the name is said.

Credits to Michel van Vliet for the audio file. 

**What's in here**

- `website/src/components/PronunciationButton/` — a small, reusable client component. The `Audio` element is created on first click rather than on render, so nothing touches `window` during Docusaurus SSR and the file is only fetched when a visitor actually asks for it (`preload="none"`).
- `website/src/pages/index.js` / `index.module.css` — renders the button pinned to the bottom-right corner of the hero logo tile. No layout shift, and the surrounding hero is untouched.
- `website/static/audio/maester-pronunciation.mp3` — the recording. 0.96 s, mono, 45 KB, silent at both ends so playback doesn't click.
- `.gitattributes` — marks `*.mp3` binary, alongside the other binary asset types already listed.

**Details**

- Accessible: real `<button>` with `aria-label` and `title` ("Hear how Maester is pronounced"), keyboard focusable, with a visible focus ring.
- The speaker waves pulse while the clip plays; the animation is disabled under `prefers-reduced-motion`.
- Playback failures — autoplay policy, a missing file, Safari returning `undefined` from `play()` — are caught and reset the button rather than leaving it stuck or throwing.
- The component takes `src`, `label` and `className` props, so it can be dropped next to another mention of the name later (an FAQ entry, for example) without changes.

## ✅ Checks

`pester.ps1` is not applicable here — this touches no PowerShell. `npm run build` in `website/` passes locally, which is what the Build Website workflow runs for this path.

## ℹ️ Additional Information

Website-only change; no module or test code is affected. Verified in the Docusaurus dev server and with a full production build: the audio is requested only on click, decodes and plays, and the button returns to its resting state when the clip ends.

Happy to also add a "How do you pronounce Maester?" entry to the FAQ in a follow-up if that's wanted — left out here to keep this focused.

## Summary by CodeRabbit

- **New Features**
  - Added a pronunciation button beside the homepage hero logo.
  - Visitors can play an audio clip to hear how “Maester” is pronounced.
  - The button provides visual playback feedback and supports hover, focus, and reduced-motion accessibility preferences.
  - Audio loads only when requested, helping improve initial page performance.
  - Playback resets correctly when the pronunciation source changes and handles playback completion or errors gracefully.


merill (approved):
Love this. Thanks @BakkerJan

PR #2173, opened 2026-09-06
fix(MT.1024): exclude configured break-glass accounts from sign-in/user risk recommendations

### LINKED ISSUE #2103: 🪲 CISA.MS.AAD.7.5 MT.1024.signinRiskPolicy MT.1024.userRiskPolicy do not consider configured breakglass account
### Pre-submission checklist

### Describe the bug

**CISA.MS.AAD.7.5: Provisioning users to highly privileged roles SHALL NOT occur outside of a PAM system.**
is testing if assignments have a start date defined.
Within the result the test is complaining about our configured Breakglass Accounts which are permanently assigned to "Global Administrator" role - I think we are aligned with genereal breakglass recommendations there.
However, the test does not extract breakglass accounts from considerations.

**MT.1024.signinRiskPolicy: Protect all users with a sign-in risk policy.**
**MT.1024.userRiskPolicy: Protect all users with a user risk policy .**
are assuming sign-in / user-risk risk policy for _all_ accounts, not excepting breakglass.

### Steps to reproduce

1. Have a breakglass account configured
2. Assign breakglass account permanently to Global Administrator role in Entra
3. Run CISA.MS.AAD.7.5 and check result --> see failed report with breakglass accounts in result
4. Have a CA with sign-risk for all users and except breakglass accounts
5. Run MT.1024 --> see failed report with breakglass accounts in results

### Expected behavior

Configured breakglass accounts shall be extracted from the result.
They might be noted informationally, but should not influence the test result negatively.

### Debug output

```text

```

### Module version

```text
ModuleType Version    PreRelease Name                                PSEdition ExportedCommands
---------- -------    ---------- ----                                --------- ----------------
Script     2.2.0                 Maester                             Core,Desk {Add-MtMaesterAppFederatedCredential, A…
```

### Environment data

```text
PSVersion                      7.6.4
GitCommitId                    7.6.4
OS                             Microsoft Windows 10.0.26200
Platform                       Win32NT
PSRemotingProtocolVersion      2.4
```

### Screenshots

Image

Image

Image

### Additional context

_No response_

earbona23:
Fixes #2103.

## The problem

`MT.1024.signinRiskPolicy` and `MT.1024.userRiskPolicy` are driven by the Entra recommendations returned from `directory/recommendations`. The check asserts the recommendation `status` is `completedBySystem`:

```powershell
$_.status | Should -Be "completedBySystem" -Because $_.benefits
```

These two recommendations flag **every** account that is not protected by a risk-based policy. Emergency access (break-glass) accounts are intentionally excluded from risk-based Conditional Access policies — that is the standard recommendation, precisely so the accounts needed to recover the tenant are never locked out by a risk detection. As a result they permanently surface as impacted resources, the recommendation never reaches `completedBySystem`, and the check fails even though the tenant is configured exactly as intended. This is what @blindzero reported, with the break-glass accounts already declared in `EmergencyAccessAccounts` in `maester-config.json`.

## The fix

For the two break-glass-relevant recommendation types only (`signinRiskPolicy`, `userRiskPolicy`), before the assertion the check now compares the still-open impacted resources against the configured emergency access accounts. If **every** impacted resource still requiring action maps to a configured break-glass account, the recommendation is treated as passed and the break-glass accounts are listed informationally in the result detail — matching the reporter's expectation that they "might be noted informationally, but should not influence the test result negatively."

The break-glass accounts are read via the existing internal `Get-MtEmergencyAccessAccount` helper (the same `EmergencyAccessAccounts` source and resolution used by `Test-MtGsaCompliantNetworkBreakGlassExcluded`), so no new configuration surface is introduced. Impacted resources are matched by `subjectId` (which maps to the affected user's object id) with `id` as a fallback.

This follows the break-glass handling precedent established in #2165 (MT.1191): reuse the configured emergency-access source, and never let the break-glass path turn into a false pass.

## Safety / what still fails

- A **real** unprotected account still fails the test — the override only applies when break-glass accounts are the *sole* remaining impacted resources.
- The scope is limited to `signinRiskPolicy` / `userRiskPolicy`; every other recommendation type is evaluated exactly as before.
- If the configured break-glass accounts **cannot be resolved**, the exclusion is skipped and the recommendation falls through to the normal (failing) assertion rather than passing — no false pass.

## Tests

Adds `powershell/internal/Test-MtRecommendationBreakGlassOnly.ps1` (the decision logic) and `powershell/tests/functions/Test-MtRecommendationBreakGlassOnly.Tests.ps1` covering: no accounts configured, single/multiple break-glass-only, a non-break-glass account still open, `completedBySystem` resources ignored, matching by `id` fallback, and a `$null` collection.

The new tests were validated by mutation — reverting the `.Count -eq 0` decision to `return $true` fails 2 of them, and dropping the `completedBySystem` filter fails 1 — so they fail if the fix is removed.

Full local suites after the change (Pester 6.1.0, pwsh 7.6.4, Linux):
- `tests/functions`: **4108 passed, 0 failed**, 800 skipped
- `tests/general`: **1186 passed, 0 failed**
- PSScriptAnalyzer suite: **987 passed, 0 failed**

The MT.1024 doc is generated by `website/scripts/generate-test-docs.mjs`, so it is not edited by hand.

## Out of scope

The issue also mentions `CISA.MS.AAD.7.5` (PIM assignment start date flagging break-glass). That is a separate check with a different data model and is intentionally left for a follow-up.

## Summary by CodeRabbit

* **New Features**
  * Entra recommendation checks now recognize configured emergency access accounts for user and sign-in risk policies.
  * Recommendations pass when unresolved impacted resources consist only of those emergency access accounts, with an informational note identifying them.
  * Resources already completed by the system continue to be excluded from evaluation.

* **Bug Fixes**
  * Recommendations are evaluated normally when emergency access accounts cannot be identified, preventing incorrect passes.

* **Tests**
  * Added coverage for emergency access matching, unresolved resources, completed resources, and missing account data.

### DIFF
--- a/powershell/internal/Test-MtRecommendationBreakGlassOnly.ps1
+++ b/powershell/internal/Test-MtRecommendationBreakGlassOnly.ps1
@@ -1,0 +1,57 @@
+﻿function Test-MtRecommendationBreakGlassOnly {
+    <#
+    .SYNOPSIS
+        Determines whether the only unresolved impacted resources of an Entra recommendation are configured emergency access (break-glass) accounts.
+
+    .DESCRIPTION
+        The sign-in risk and user risk recommendations (MT.1024.signinRiskPolicy / MT.1024.userRiskPolicy)
+        flag every account that is not protected by a risk-based policy. Emergency access (break-glass)
+        accounts are intentionally excluded from risk-based Conditional Access policies to avoid locking
+        out the accounts needed to recover the tenant, so they always surface as impacted resources that
+        will never be remediated.
+
+        This function returns $true only when there is at least one impacted resource still requiring
+        action and every such resource maps to a configured break-glass account (matched by subjectId or
+        id). In that case the break-glass accounts are the sole reason the recommendation is not complete
+        and it must not fail the test.
+
+        It returns $false when at least one non break-glass resource still requires action, when nothing
+        is still open, or when no break-glass object IDs are supplied (nothing to exclude).
+
+    .EXAMPLE
+        Test-MtRecommendationBreakGlassOnly -ImpactedResources $recommendation.impactedResources -BreakGlassObjectId $ids
+
+        Returns $true if the only impacted resources still requiring action are the configured break-glass accounts.
+    #>
+    [CmdletBinding()]
+    [OutputType([bool])]
+    param (
+        # The impactedResources collection of an Entra recommendation.
+        [Parameter(Mandatory)]
+        [AllowNull()]
+        [AllowEmptyCollection()]
+        $ImpactedResources,
+
+        # Directory object IDs of the configured emergency access (break-glass) accounts.
+        [Parameter(Mandatory)]
+        [AllowNull()]
+        [AllowEmptyCollection()]
+        [string[]] $BreakGlassObjectId
+    )
+
+    # Nothing configured to exclude - the recommendation must be evaluated as-is.
+    if (-not $BreakGlassObjectId) { return $false }
+
+    # Only resources the system has not already resolved still count against the recommendation.
+    $openResources = @($ImpactedResources | Where-Object { $_.status -ne 'completedBySystem' })
+    if ($openResources.Count -eq 0) { return $false }
+
+    # subjectId maps to the affected object (a user object id for the risk recommendations); id is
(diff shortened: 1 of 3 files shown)


merill (approved):
This is an elegant fix for the issue @earbona23 

Thanks a lot!

PR #2221, opened 2026-09-20
fix: reuse EAM classification for PIM alerts

### LINKED ISSUE #2218: Reuse pinned EAM classification for Test-MtPimAlertsExists
When `-FilteredAccessLevel` is supplied, `Test-MtPimAlertsExists` downloads the same mutable [EntraOps classification](https://github.com/Cloud-Architekt/AzurePrivilegedIAM/blob/main/Classification/Classification_EntraIdDirectoryRoles.json) inside `process`, repeating the download for pipeline items.

Reuse the checked-in table from #2217, loading it on the first filtered item and reusing it for the invocation. Apply each item's tier filter after pipeline properties are bound; `begin` is too early for `ValueFromPipelineByPropertyName`.

Preserve the existing error/skip behavior when a requested filter cannot be applied. Although the filter is optional to callers, silently widening a requested scope can change the result of a tier-specific check. Cover direct calls, changing pipeline tiers, empty/missing classification, and absence of runtime web requests.

Implementation: #2221, dependent on #2220.

nathanmcnulty:
Fixes #2218. Depends on #2220; [review just the PIM changes](https://github.com/nathanmcnulty/maester/compare/fix/embed-eam-classification...fix/pim-alerts-classification). The shared snapshot changes are included until that PR merges.

Replace the per-item GitHub download with the checked-in EAM table. Load it on the first filtered item, cache it for the invocation, and derive each item's filter after pipeline property binding.

Preserve existing error/skip behavior if a requested filter cannot be applied. Returning broader results would change the scope of a tier-specific check.

Validation: 8 PIM tests pass, covering direct filtering, changing pipeline tiers, one classification load, empty/missing classification, and no runtime web requests.


merill (approved):
LGTM

